PR #2, opened 2019-06-13
Add support for RSHIM under FreeBSD.

hselasky:


Makefile
```diff
@@ -15,6 +15,7 @@ CDEF = -D_FILE_OFFSET_BITS=64 -DRSHIM_LOG_ENABLE=1 -DFUSE_USE_VERSION=30
 CDEF += -DHAVE_RSHIM_NET
 CDEF += -DHAVE_RSHIM_PCIE
 CDEF += -DHAVE_RSHIM_PCIE_LF
+CDEF += -DHAVE_RSHIM_FUSE
```
lsun100: Use "HAVE_RSHIM_FUSE" for Linux and "HAVE_RSHIM_CUSE" for FreeBSD seems a little confusing. 
Linux code actually uses both fuse and cuse. Probably use something else for this difference?

rshim.c
```diff
@@ -2132,7 +2799,13 @@ static int rshim_fs_init(struct rshim_backend *bd)
   char *argv[] = {"./bfrshim", "-f"};
   int i, rc;
 
+#ifdef HAVE_RSHIM_CUSE
+  if (cuse_init() != CUSE_ERR_NONE)
+    return -1;
+#endif
+
   for (i = 0; i < RSH_DEV_TYPES; i++) {
+#ifdef FUSE
```
lsun100: Is it a typo here?  Linux doesn't have this definition.

hselasky: Yes, this is a typo.

hselasky: should be fixed now.

Makefile
hselasky: ifdef __linux__ and ifdef __FreeBSD__ is not the right thing for this. Do you have any suggestions?

lsun100: I don't have good suggestions either.. Let's make it working first. We could consider improving it later, such as isolating the fs implementation into separate files...

rshim.c
lsun100: Other code looks ok to me. I'll try to verify it in Linux today.

PR #13, opened 2020-03-12
Split fuse code into rshim_fuse.c

lsun100:
This commit splits the fuse/cuse code into rshim_fuse.c. So fuseis just one way to implement the boot/console/networking services. Other user interfaces, like networking or CLI, could be added
when needed.

This commit also unifies the build process for Linux and FreeBSD, and adds '--enable-fuse' in the configure options.


hselasky:
Looks good to me.

PR #23, opened 2020-07-24
Add configuration file support

lsun100:
This commit adds configuration file support (/etc/rshim.conf)
which can be used to configure the static mapping of rshim
device to rshim name as well as the rshim devices to be ignored.

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -173,9 +173,13 @@ int rshim_epoll_fd;
 int rshim_static_index = -1;
 char *rshim_static_dev_name;
 
+/* Default configuration file. */
+char *rshim_cfg_file = "/etc/rshim.conf";
+
 /* Array of devices and device names. */
 rshim_backend_t *rshim_devs[RSHIM_MAX_DEV];
 char *rshim_dev_names[RSHIM_MAX_DEV];
+char *rshim_blocked_dev_names[RSHIM_MAX_DEV];
 
 int rshim_log_level = LOG_NOTICE;
 bool rshim_daemon_mode = true;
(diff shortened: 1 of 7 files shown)


src/rshim.c
```diff
@@ -2376,6 +2390,56 @@ int rshim_fifo_size(rshim_backend_t *bd, int chan, bool is_rx)
...
+
+  while (getline(&buf, &n, file) != -1) {
+    if (sscanf(buf, "%31s%63s", rshim_name, dev_name) != 2)
+      continue;
+
+    if (strncmp(rshim_name, "rshim", 5) && strcmp(rshim_name, "none"))
+      continue;
+
+    if (strncmp(dev_name, "usb-", 4) && strncmp(dev_name, "pcie-", 5) &&
+        strncmp(dev_name, "pcie-lf", 7))
```
shravankumarr, line 2413: checking for "pcie-" should cover both "pcie-" and "pcie-lf" cases. so the last check isn't necessary

lsun100: good catch. done.

src/rshim.c
```diff
@@ -2376,6 +2390,55 @@ int rshim_fifo_size(rshim_backend_t *bd, int chan, bool is_rx)
...
+
+  file = fopen(rshim_cfg_file, "r");
+  if (!file)
+    return -ENOENT;
+
+  while (getline(&buf, &n, file) != -1) {
+    if (sscanf(buf, "%31s%63s", rshim_name, dev_name) != 2)
+      continue;
+
+    if (strncmp(rshim_name, "rshim", 5) && strcmp(rshim_name, "none"))
```
shravankumarr, line 2409: sorry, missed this. please use strncmp for the "none" check also

PR #25, opened 2020-07-29
Display/configure OPN string for BlueField-2

lsun100:
This commit supports the display and configuring of the OPN string
for BlueField-2 cards. The setting is in the YU boot record which
persists during warm reboot and will be lost after cold reboot.

### DIFF
--- a/src/rshim_fuse.c
+++ b/src/rshim_fuse.c
@@ -613,6 +613,7 @@ static int rshim_fuse_misc_read(struct cuse_dev *cdev, int fflags,
   struct rshim_misc *rm = cuse_dev_get_per_file_handle(cdev);
   off_t off;
 #endif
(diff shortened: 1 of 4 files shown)


src/rshim_fuse.c
```diff
@@ -893,6 +907,10 @@ static int rshim_fuse_misc_write(struct cuse_dev *cdev, int fflags,
     bd->has_cons_work = 1;
     rshim_work_signal(bd);
     pthread_mutex_unlock(&bd->mutex);
+  } else if (!strcmp(key, "OPN_STR")) {
+    if (sscanf(p, "%16s", opn) != 1)
```
shravankumarr, line 911: don't we need a check here for RSHIM_BLUEFIELD_2 cuz this isn't supported in BF1?

PR #28, opened 2020-08-13
Add workaround to boot Viper rev A0 in LiveFish mode

dwoods2:
BF-2 rev A0 requires a 'type 13' reset following each soft reset
in order to avoid a mesh deadlock.  This requires configuring
the type 13 reset modes at attach time and then after each soft
reset also write the reset 13 activation bit.

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -503,10 +503,11 @@ static int rshim_reg_indirect_wait(rshim_backend_t *bd, uint64_t resp_count)
     if (count != resp_count)
       return 0;
   }
+  RSHIM_ERR("Rshim byte access widget timeout\n");
   return -1;
 }
 
-static void rshim_mmio_write_common(rshim_backend_t *bd, uintptr_t pa,
+static int rshim_mmio_write_common(rshim_backend_t *bd, uintptr_t pa,
                                     uint8_t size, uint64_t data)
 {
   uint64_t reg, resp_count;
@@ -524,12 +525,17 @@ static void rshim_mmio_write_common(rshim_backend_t *bd, uintptr_t pa,
         (1ULL << RSH_MEM_ACC_CTL__WRITE_SHIFT) |
         (1ULL << RSH_MEM_ACC_CTL__SEND_SHIFT);
   bd->write_rshim(bd, RSHIM_CHANNEL, RSH_MEM_ACC_CTL, reg);
-  rshim_reg_indirect_wait(bd, resp_count);
+  return rshim_reg_indirect_wait(bd, resp_count);
 }
 
-uint64_t rshim_mmio_read_common(rshim_backend_t *bd, uintptr_t pa, uint8_t size)
+static int rshim_mmio_read_common(rshim_backend_t *bd, uintptr_t pa,
+                                  uint8_t size, uint64_t *data)
 {
   uint64_t reg, resp_count;
+
+  bd->read_rshim(bd, RSHIM_CHANNEL, RSH_DEVICE_MSTR_PRIV_LVL, &reg);
+  reg |= 0x1ULL << RSH_DEVICE_MSTR_PRIV_LVL__MEM_ACC_LVL_SHIFT;
+  bd->write_rshim(bd, RSHIM_CHANNEL, RSH_DEVICE_MSTR_PRIV_LVL, reg);
 
   bd->read_rshim(bd, RSHIM_CHANNEL, RSH_MEM_ACC_RSP_CNT, &resp_count);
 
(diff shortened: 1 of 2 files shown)


src/rshim.c
```diff
@@ -1989,6 +2028,82 @@ static void rshim_boot_workaround_check(rshim_backend_t *bd)
...
 }
 
+static int rshim_bf2_a0_wa(rshim_backend_t *bd)
+{
+  int i, rc;
+  uint32_t devid, clk_en, reset_en, powerdown;
+  uint32_t main_clk_gate_en;
+
+  rc = rshim_mmio_read32(bd, RSHIM_YU_BASE_ADDR + YU_BOOT_DEVID, &devid);
+  RSHIM_INFO("yu_boot_devid: 0x%x\n", devid);
```
lsun100, line 2038: RSHIM_DBG()?   this log message will be in Linux syslog if printed.

src/rshim.c
```diff
@@ -1989,6 +2028,82 @@ static void rshim_boot_workaround_check(rshim_backend_t *bd)
   }
 }
 
+static int rshim_bf2_a0_wa(rshim_backend_t *bd)
+{
+  int i, rc;
```
lsun100, line 2033: Maybe move this line after line 2035?  The code tries to follow the 'mellanox' way to sort the lines in descending according to the line width.

src/rshim.c
```diff
@@ -524,13 +525,18 @@ static void rshim_mmio_write_common(rshim_backend_t *bd, uintptr_t pa,
...
 
-uint64_t rshim_mmio_read_common(rshim_backend_t *bd, uintptr_t pa, uint8_t size)
+uint64_t rshim_mmio_read_common(rshim_backend_t *bd, uintptr_t pa,
+                                uint8_t size, uint64_t *data)
 {
   uint64_t reg, resp_count;
 
+  bd->read_rshim(bd, RSHIM_CHANNEL, RSH_DEVICE_MSTR_PRIV_LVL, &reg);
+  reg |= 0x1ULL << RSH_DEVICE_MSTR_PRIV_LVL__MEM_ACC_LVL_SHIFT;
+  bd->write_rshim(bd, RSHIM_CHANNEL, RSH_DEVICE_MSTR_PRIV_LVL, reg);
```
lsun100, line 538: Add a check to only do this for BF2 (since BF1 is different)

src/rshim.c
```diff
@@ -524,13 +525,18 @@ static void rshim_mmio_write_common(rshim_backend_t *bd, uintptr_t pa,
         (1ULL << RSH_MEM_ACC_CTL__WRITE_SHIFT) |
         (1ULL << RSH_MEM_ACC_CTL__SEND_SHIFT);
   bd->write_rshim(bd, RSHIM_CHANNEL, RSH_MEM_ACC_CTL, reg);
-  rshim_reg_indirect_wait(bd, resp_count);
+  return rshim_reg_indirect_wait(bd, resp_count);
 }
 
-uint64_t rshim_mmio_read_common(rshim_backend_t *bd, uintptr_t pa, uint8_t size)
+uint64_t rshim_mmio_read_common(rshim_backend_t *bd, uintptr_t pa,
```
lsun100, line 531: uint64_t ==> int

src/rshim.c
```diff
@@ -593,6 +625,13 @@ int rshim_reset_control(rshim_backend_t *bd)
     return rc;
   }
 
+  if (bd->ver_id == RSHIM_BLUEFIELD_2 && rshim_is_livefish(bd)) {
+    RSHIM_INFO("Apply reset type 13\n");
+    /* yu.reset_mode_control.reset_mode_control.sw_reset_event_activation13 */
+    rshim_mmio_write32(bd, RSHIM_YU_BASE_ADDR + YU_RESET_ACTIVATION_13, 1);
+
```
lsun100, line 632: remove this extra line

src/rshim.c
```diff
@@ -540,22 +546,48 @@ uint64_t rshim_mmio_read_common(rshim_backend_t *bd, uintptr_t pa, uint8_t size)
...
+  uint64_t priv_lvl;
+  int rc;
+
+  /*
+   * A value of 1 in yu_boot.boot_status indicates a successful FW
+   * boot, any other value indicates livefish mode.
+   */
+  rc = rshim_mmio_read32(bd, RSHIM_YU_BASE_ADDR + YU_BOOT, &yu_boot);
+  boot_status = (yu_boot >> 17) & 3;
+  RSHIM_INFO("yu_boot_status: %d\n", boot_status);
```
lsun100, line 588: RSHIM_DBG() ?

src/rshim.c
dwoods2: Not sure I understand this one.  The mem acc widget security level setting is the same between BF1/BF2.  The name of the register changed, but the location and meaning are the same.

lsun100: Please ignore this comment. Yes, the offset and bit are the same for both BF1 anf BF2.

Issue #26, opened 2020-08-12
rshim should exit when cuse is missing


mgerdts:
rshim has no chance of doing anything useful when the `cuse` kernel module cannot be loaded.  It should log an appropriate message and exit non-zero immediately.  This would make debugging this failure mode much easier.

This is with rshim built from commit 9b1da8407833cab4966c5f110df7683b646c2489.

```
# ./rshim -b pcie -d 0000:d8:00.2 -l 4 -f
modprobe: FATAL: Module cuse not found in directory /lib/modules/4.18.0-80.el8.27782638.x86_64
```

At this point it just keeps running.

hselasky:
Which OS are you using? cuse.ko is only for FreeBSD ?

mgerdts:
I'm using CentOS 8 with an in-house kernel.  I have subsequently found that the `cuse` module comes from `kernel-modules-extra`, which our internal build process has inconveniently not propagated beyond build servers.  That's something that I need to rectify.

That finding suggest that the `rshim` rpm could benefit from a dependency on `kernel-modules-extra` to head off this problem.

lsun100:
The rshim.spec file does have check for kernel-modules-extra:
%if 0%{?rhel} >= 8 || 0%{?fedora} > 0
Requires: kernel-modules-extra
%endif

Probably the check failed. I'll take to take a look.

lsun100:
Mike, the RPM does check the kernel-modules-extra (see below). Probably the RPM binary you used was built in a slightly different way. The source RPM in the release tarball should be good. 

# cat /etc/redhat-release
CentOS Linux release 8.2.2004 (Core)

#rpmbuild --rebuild rshim-2.0.5-2.g9b1da84.src.rpm

# rpm -ivh rshim-2.0.5-2.g9b1da84.el8.x86_64.rpm
error: Failed dependencies:
        kernel-modules-extra is needed by rshim-2.0.5-2.g9b1da84.el8.x86_64

mgerdts:
Hmm, the version I have is indeed busted and different.

```
[root@xxx ~]# cat /etc/os-release
NAME="CentOS Linux"
VERSION="8 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="8"
PLATFORM_ID="platform:el8"
PRETTY_NAME="CentOS Linux 8 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:8"
[...]

PR #29, opened 2020-08-13
Quit execution if cuse is not found

lsun100:
Current rshim driver implementation relies on the fuse/cuse kernel
module in Linux. This commit quits the execution if cuse is not loaded.

Issue #26


lsun100:
#26 is already closed. no need for this change any more.

PR #31, opened 2020-08-18
Fix return value if pcie or pcie_lf device not found

mbgg:
If we are not able to find the pcie/pcie_lf device, we should return an error, so that rshim can catch that and abort.

Signed-off-by: Matthias Brugger <mbrugger@suse.com>

### DIFF
--- a/src/rshim_pcie.c
+++ b/src/rshim_pcie.c
@@ -529,6 +529,7 @@ int rshim_pcie_init(void)
   struct pci_access *pci;
   struct pci_dev *dev;
   int rc;
+  bool dev_present = false;
 
   pci = pci_alloc();
   if (!pci)
@@ -552,7 +553,11 @@ int rshim_pcie_init(void)
       continue;
 
     rshim_pcie_probe(dev);
+    dev_present = true;
   }
+
+  if (!dev_present)
+	  return -1;
 
   return 0;
 }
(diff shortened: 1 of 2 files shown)


src/rshim_pcie.c
```diff
@@ -554,5 +554,8 @@ int rshim_pcie_init(void)
     rshim_pcie_probe(dev);
   }
 
+  if (!dev)
+	  return -1;
+
```
lsun100, line 559: The above loop doesn't return. So it'll always come here with dev equals to NULL. Probably you meant if no pcie device was found in the above loop then return -1?

src/rshim_pcie_lf.c
```diff
@@ -618,5 +618,8 @@ int rshim_pcie_lf_init(void)
     rshim_pcie_probe(dev);
   }
 
+  if (!dev)
+	  return -1;
+
```
lsun100, line 623: same comment as above.

src/rshim_pcie.c
mbgg: Right, stupid me. I'll push a new version setting a boolean when we find the first device.

mbgg:
Sorry I'm not that familiar with the github workflow, I just force pushed the branch. Let me know if you prefer fixes of the commit that you will squash when accepting. Then I can do that in the future.

PR #50, opened 2021-01-27
Fix potential tmfifo data loss

lsun100:
The tmfifo output was stored in a ring buffer. When it's time to
transmit, it calls rshim_fifo_output() to pack data into a buffer,
then calls the backend->write() to write to the fifo which calls
rshim_write_default() for pcie rshim. The buffer will eventually
calls rshim_write_delayed() to process it which checks the available
space before writting to the fifo, and will timeout if no space
in 3 seconds. Once it happens, it'll cause data loss and fifo packet
out-of-sync. It could happen when ARM is too slow (such as during
booting) under external tmfifo network traffic.

This commit fixes the problem by checking the available space in
rshim_fifo_output() when preparing the write buffer. So it won't
have any timeout later.

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -32,6 +32,9 @@
 /* Cycles to poll the network initialization before timeout. */
 #define RSHIM_NET_INIT_DELAY (60000 / RSHIM_TIMER_INTERVAL)
 
+/* Reserve some space to indicate full. */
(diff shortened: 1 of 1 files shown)


lsun100:
Shravan might be still on vacation. Need to get it in today to catch a release build.

lsun100:
Merge it for now. Will handle Shravan's comments later.

PR #51, opened 2021-02-06
rshim_usb.c: Make necessary changes for Mustang

asmaamellanox:
There are no differences in the registers.
The following are the changes made:
1) BF-3 product ID, this is just a place holder.
Still waiting for the HW folks to provide it.
2) BF-2 Vendor ID remains the same i.e. Mellanox
3) the channel is no longer supported in BF-3.
Instead, the register address is computed as:
(wValue << 16) + wIndex


src/rshim_usb.c
```diff
@@ -75,23 +77,34 @@ static void rshim_usb_delete(rshim_backend_t *bd)
   free(dev);
 }
 
+static int chan_to_wValue(int chan, uint16_t ver_id)
+{
+  if (ver_id == RSHIM_BLUEFIELD_3)
+    return chan << 7;
+
+  return chan;
```
lsun100, line 85: We might need a table to update both wValue & wIndex for BF3 (by setting the the wValue and adding an offset to wIndex ).

src/rshim_usb.c
```diff
@@ -75,23 +77,34 @@ static void rshim_usb_delete(rshim_backend_t *bd)
   free(dev);
 }
 
+static int chan_to_wValue(int chan, uint16_t ver_id)
```
lsun100, line 80: It's ok to use lower case as function name to be consistent.

lsun100 (commented):
Could you also add "Signed-off-by:" just like other commits?

src/rshim_usb.c
asmaamellanox: done.

src/rshim_usb.c
asmaamellanox: done

asmaamellanox:
done

lsun100 (approved):

PR #53, opened 2021-03-08
Reduce CPU utilization when there is no rshim device

lsun100:
This commit checks the current attached rshim devices and uses a
large interval for the background timer when there is no rshim
devices. Thus the CPU utilization will be minimum in such case.

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -28,6 +28,7 @@
 
 /* RShim timer interval in milliseconds. */
 #define RSHIM_TIMER_INTERVAL 1
+#define RSHIM_TIMER_INTERVAL_SLOW 100
 
 /* Cycles to poll the network initialization before timeout. */
 #define RSHIM_NET_INIT_DELAY (60000 / RSHIM_TIMER_INTERVAL)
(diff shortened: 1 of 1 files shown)


src/rshim.c
```diff
@@ -187,11 +188,19 @@ rshim_backend_t *rshim_devs[RSHIM_MAX_DEV];
 char *rshim_dev_names[RSHIM_MAX_DEV];
 char *rshim_blocked_dev_names[RSHIM_MAX_DEV];
 
+/* Bitmask of the used rshim device id. */
+#if RSHIM_MAX_DEV > 64
```
asmaamellanox, line 192: DId you add this to fix it in the future?
Since it is set to 64 right now?

lsun100: This commit uses a 64-bit bitmask to identify the connected rshim devices. It's based on the assumption that the RSHIM_MAX_DEV won't be bigger than 64.

This line is a robust check to prevent us increasing RSHIM_MAX_DEV larger than 64 by accident.

(64 should be good enough. If more rshim devices are needed someday, we could split them into different processes).

asmaamellanox (approved):

PR #69, opened 2022-01-25
Mustang: Add support for rshim over pcie and pcie_lf

shravankumarr:
Use MSN_GW path to access rshim in livefish mode.
rshim_pcie uses the 8M BAR to access the rshim.


src/rshim_pcie.c
```diff
@@ -270,8 +278,13 @@ rshim_pcie_read(rshim_backend_t *bd, int chan, int addr, uint64_t *result)
...
 
 #ifndef __LP64__
+  if (rshim_is_bluefield3(dev->pci_dev->device_id))
+    return -EIO;
+
   rc = rshim_byte_acc_read(dev, RSH_CHANNEL_BASE(chan) + addr, result);
 #else
+  if (rshim_is_bluefield3(dev->pci_dev->device_id))
+    chan = 0;
   *result = readq(dev->rshim_regs + (addr | (chan << 16)));
```
lsun100, line 288: I think we might need a channel map here for PCIE like below.  Or else it can only access the first 64K address with this BAR mapping.

if ((rshim_is_bluefield3(dev->pci_dev->device_id)) {
    *result = readq(dev->rshim_regs + bf3_rshim_pcie_chan_map[chan])
} else {
    *result = readq(dev->rshim_regs + (addr | (chan << 16)));
}

The bf3_rshim_channel_map will have some definition like below, similar to bf3_wvalue_widx_pair_map with adjusted offset.

uint32_t bf3_rshim_pcie_chan_map[] = {
    [RSHIM_CHANNEL] = 0;
    [UART0_CHANNEL] = ...;
};

src/rshim_pcie_lf.c
```diff
@@ -93,6 +109,201 @@ static int pci_cap_write(struct pci_dev *pci_dev, int offset, uint32_t value)
...
+
+  /* Acquire MSN_GW_NIC_FW_LOCK */
+  rc = msn_gw_lock_acquire(pci_dev);
+  if (rc)
+    return rc;
+
+  /* Write addr to MSN_GW_ADDR */
+  rc = pci_cap_write(pci_dev, MSN_GW_CR_64B_NIC_FW_ADDR, (addr >> 2));
+  if (rc)
+    return rc;
```
lsun100, line 184: We should release the lock for all the cases, such as using goto here and release the lock at the end.

src/rshim_pcie_lf.c
```diff
@@ -93,6 +109,201 @@ static int pci_cap_write(struct pci_dev *pci_dev, int offset, uint32_t value)
...
+  /* Trigger MSN_GW to read from addr */
+  rc = pci_cap_read(pci_dev, MSN_GW_CR_64B_NIC_FW_REG0, &data);
+  if (rc)
+    return rc;
+
+  data |= MSN_GW_CR_64B_NIC_FW_BUSY;
+
+  rc = pci_cap_write(pci_dev, MSN_GW_CR_64B_NIC_FW_REG0, data);
+  if (rc)
+    return rc;
```
lsun100, line 206: The logic of these several lines are a little suspicious. Should we do line 202-204 first if line 198 is the actual trigger?

src/rshim_pcie_lf.c
```diff
@@ -93,6 +109,201 @@ static int pci_cap_write(struct pci_dev *pci_dev, int offset, uint32_t value)
...
+  /* Trigger MSN_GW to write value to addr */
+  rc = pci_cap_read(pci_dev, MSN_GW_CR_64B_NIC_FW_REG0, &data);
+  if (rc)
+    return rc;
+
+  data |= MSN_GW_CR_64B_NIC_FW_BUSY;
+
+  rc = pci_cap_write(pci_dev, MSN_GW_CR_64B_NIC_FW_REG0, data);
+  if (rc)
+    return rc;
```
lsun100, line 283: same question here

src/rshim_pcie_lf.c
```diff
@@ -93,6 +109,201 @@ static int pci_cap_write(struct pci_dev *pci_dev, int offset, uint32_t value)
...
+
+  /* Acquire MSN_GW_NIC_FW_LOCK */
+  rc = msn_gw_lock_acquire(pci_dev);
+  if (rc)
+    return rc;
+
+  /* Write addr to MSN_GW_ADDR */
+  rc = pci_cap_write(pci_dev, MSN_GW_CR_64B_NIC_FW_ADDR, (addr >> 2));
+  if (rc)
+    return rc;
```
lsun100, line 249: same here, release the lock for all cases

src/rshim_pcie_lf.c
shravankumarr: I'm doing a RMW to the reg setting only the BUSY bit to trigger the action. The comment is misleading, updated.

src/rshim_pcie.c
shravankumarr: Do you think that would be useful given that we will only be allowed to access the registers with "BOOT" privilege in this path?

src/rshim_pcie_lf.c
shravankumarr: Done

src/rshim_pcie.c
lsun100: I see. Is it the mapping we added in BL1?

shravankumarr: Yes, we program the PRNF_DMA_CRWIN in BL1 to map this.

lsun100 (approved):

PR #72, opened 2022-02-16
PCIe: VFIO support

lsun100:
This commits adds VFIO support for PCIe backend in order to support
the secure-boot host where kernel lock-down is enable thus direct
mmap() of the sysfs resource file is not allowed.

This commit also improves the 'DROP_MODE' handling. When DROP_MODE
is set, the PCIe resource will be released so the rshim PF could be
re-used for other purpose at runtime such as passing-through to other
VMs.

### DIFF
--- a/src/rshim_pcie.c
+++ b/src/rshim_pcie.c
@@ -4,20 +4,33 @@
  *
  */
 
-#include <sys/epoll.h>
-#include <sys/mman.h>
+#include <errno.h>
+#include <fcntl.h>
 #include <pci/pci.h>
+#include <pthread.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/epoll.h>
+#include <sys/ioctl.h>
+#include <sys/mman.h>
+#include <sys/param.h>
 #include <sys/poll.h>
 #include <sys/socket.h>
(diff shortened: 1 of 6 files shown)


src/rshim_pcie.c
```diff
@@ -34,10 +47,18 @@
...
-#define SYS_BUS_PCI "/sys/bus/pci/devices"
-#endif
+#define SYS_CLASS_IOMMU_PATH        "/sys/class/iommu"
+#define SYS_CLASS_VFIO_PCI_PATH     "/sys/module/vfio_pci"
+#define SYS_BUS_PCI_PATH            "/sys/bus/pci/devices"
+#define SYS_VFIO_PCI_PATH           "/sys/bus/pci/drivers/vfio-pci"
+
+typedef enum {
+  RSHIM_PCIE_MMAP_DIRECT,
+  RSHIM_PCIE_MMAP_UIO,
```
shravankumarr, line 57: Is UIO the same as mmap_direct?

lsun100: UIO is not implemented yet and will be added as part of the interrupt support for April release.
It'll be similar to VFIO but much simpler.

shravankumarr (approved):

PR #73, opened 2022-03-03
PCIe: Remove 32-bit support

lsun100:
This commit removes 32-bit support for simplicity since the
RSH_BYTE_ACC_xxx registers have been removed from BF3 thus needs
to be reworked anyway. It can be added back when needed.


hselasky:
Just make sure the code still compiles on 32-bit platforms.

lsun100:
> 

Yes, the change can compile on freebsd-13.0-i386.

shravankumarr (approved):

PR #81, opened 2022-05-02
Fix several compiling issues for FreeBSD

lsun100:

### DIFF
--- a/src/rshim_pcie.c
+++ b/src/rshim_pcie.c
@@ -12,7 +12,6 @@
 #include <stdio.h>
 #include <string.h>
 #include <sys/epoll.h>
-#include <sys/eventfd.h>
 #include <sys/ioctl.h>
 #include <sys/mman.h>
 #include <sys/param.h>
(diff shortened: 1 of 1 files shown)


src/rshim_pcie.c
```diff
@@ -192,7 +191,7 @@ typedef struct {
 int bf3_rshim_pcie_chan_map[] = {
```
hselasky, line 191: BTW: Should this array be "static const int bf3_rshim_pcie_chan_map[]" ?

hselasky:
Changes look good!

src/rshim_pcie.c
lsun100: Done. Updated version 2.

hselasky (approved):
Looks good. Just make a new tag aswell and I'll get it into the FreeBSD ports tree!

lsun100:
> Looks good. Just make a new tag as well and I'll get it into the FreeBSD ports tree!

Thanks. Done and tagged as rshim-2.0.6-10

lsun100:
Thanks, done. I re-created rshim-2.0.6-10.

hselasky:
Now upstream FreeBSD:
https://cgit.freebsd.org/ports/commit/?id=66be6b83209a10aa59cc5576ac9144f2c0a44f51

lsun100:
Nice!

PR #83, opened 2022-06-03
Detect new USB/rshim hot plugin

lsun100:
This commit enables USB probing even there is no USB/rshim devices.
In such case, blocking call is used to avoid polling.

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -2612,9 +2612,6 @@ static void rshim_main(int argc, char *argv[])
       }
     }
 
-    /* Check USB for timeout or unhandled fd. */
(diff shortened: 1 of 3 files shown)


asmaamellanox (approved):

hselasky (commented):
Looks good.

In FreeBSD the hot-plug framework runs from a separate serialized thread.

hselasky:
Make sure you test this on FreeBSD too, that it doesn't go into a spinning loop.

lsun100:
> Make sure you test this on FreeBSD too, that it doesn't go into a spinning loop.

Thanks. Yes, will check it before merging.

PR #87, opened 2022-07-13
BF3: Support 32-bit CR-space access via USB

lsun100:
This commit enhances the channel/address to access the 32-bit
CR-space via USB for BF3. When channel is bigger than 0xF, it'll
be interpreted as upper 16-bits of the linear CR-space address.

### DIFF
--- a/src/rshim_usb.c
+++ b/src/rshim_usb.c
@@ -140,8 +140,13 @@ static struct rshim_usb_addr get_wvalue_windex(int chan, int addr, uint16_t ver_
   struct rshim_usb_addr rsh_usb_addr;
 
   if (ver_id == RSHIM_BLUEFIELD_3) {
-    rsh_usb_addr.wvalue = bf3_wvalue_widx_pair_map[chan].wvalue + BF_MMIO_BASE;
-    rsh_usb_addr.windex = bf3_wvalue_widx_pair_map[chan].windex + addr;
+    if (chan <= 0xF) {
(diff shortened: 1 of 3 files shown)


src/rshim_usb.c
```diff
@@ -140,8 +140,13 @@ static struct rshim_usb_addr get_wvalue_windex(int chan, int addr, uint16_t ver_
   struct rshim_usb_addr rsh_usb_addr;
 
   if (ver_id == RSHIM_BLUEFIELD_3) {
-    rsh_usb_addr.wvalue = bf3_wvalue_widx_pair_map[chan].wvalue + BF_MMIO_BASE;
-    rsh_usb_addr.windex = bf3_wvalue_widx_pair_map[chan].windex + addr;
+    if (chan <= 0xF) {
+      rsh_usb_addr.wvalue = bf3_wvalue_widx_pair_map[chan].wvalue + BF_MMIO_BASE;
+      rsh_usb_addr.windex = bf3_wvalue_widx_pair_map[chan].windex + addr;
```
dwoods2, line 145: I don't understand the logic here.  CRspace addresses are 28 bits.  Isn't wvalue the upper 16 bits of the address?  In that case it should never be > 0xF.  Also I don't think there's any need to add in BF_MMIO_BASE since that just sets bit 28.

dwoods2 (approved):

asmaamellanox (approved):

PR #88, opened 2022-08-02
support other transfer sizes: 1B, 2B, 4B on top of 8B

asmaamellanox:
Mustang hardware supports 1B, 2B, 4B and 8B USB transfers.

### DIFF
--- a/src/rshim_fuse.c
+++ b/src/rshim_fuse.c
@@ -643,7 +643,7 @@ static int rshim_fuse_misc_read(struct cuse_dev *cdev, int fflags,
   pthread_mutex_lock(&bd->mutex);
 
   /* Boot mode. */
-  rc = bd->read_rshim(bd, RSHIM_CHANNEL, bd->regs->boot_control, &value);
+  rc = bd->read_rshim(bd, RSHIM_CHANNEL, bd->regs->boot_control, &value, RSHIM_REG_SIZE_8B);
   if (rc) {
     pthread_mutex_unlock(&bd->mutex);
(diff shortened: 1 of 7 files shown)


src/rshim_fuse.c
```diff
@@ -1002,6 +1003,7 @@ static const struct cuse_methods rshim_misc_fops = {
 typedef struct {
```
lsun100, line 1003: This change will break existing tools, such as the upstream openocd. 

Can we use a different message? Such as defining rshim_ioctl_msg2, and RSHIM_IOC_READ2, RSHIM_IOC_WRITE2.

src/rshim_fuse.c
```diff
@@ -1064,11 +1066,11 @@ static void rshim_fuse_rshim_ioctl(fuse_req_t req, int cmd, void *arg,
...
-      rc = bd->write_rshim(bd, chan, offset, msg.data);
+      rc = bd->write_rshim(bd, chan, offset, msg.data, msg.data_size);
       pthread_mutex_unlock(&bd->mutex);
     } else {
       pthread_mutex_lock(&bd->mutex);
-      rc = bd->read_rshim(bd, chan, offset, &data);
+      rc = bd->read_rshim(bd, chan, offset, &data, msg.data_size);
       msg.data = data;
       pthread_mutex_unlock(&bd->mutex);
     }
```
lsun100, line 1076: Once we have RSHIM_IOC_WRITE2/RSHIM_IOC_READ2, the code here will be slightly different

lsun100 (approved):

PR #91, opened 2022-08-10
Add configuration option to enable/disable PCIe VFIO/UIO

lsun100:
VFIO/UIO is currently enabled by default and will be selected
automatically. Some users would like to specify which one to use
or diable. This commit adds configuration option in rshim.conf
to disable them as needed by setting PCIE_HAS_VFIO / PCIE_HAS_UIO
to 0 in rshim.conf.

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -188,6 +188,9 @@ static int rshim_boot_timeout = 100;
 int rshim_drop_mode = -1;
 int rshim_usb_reset_delay = 5;
 int rshim_pcie_reset_delay = 10;
+int rshim_pcie_enable_vfio = 1;
+int rshim_pcie_enable_uio = 1;
+int rshim_pcie_intr_poll_interval = 10;  /* Interrupt polling in milliseconds */
 
 /* Array of devices and device names. */
 rshim_backend_t *rshim_devs[RSHIM_MAX_DEV];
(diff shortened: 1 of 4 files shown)


src/rshim.c
```diff
@@ -2725,17 +2728,21 @@ static int rshim_load_cfg(void)
     } else if (!strcmp(key, "DROP_MODE")) {
       rshim_drop_mode = (atoi(value) > 0) ? 1 : 0;
       continue;
-    } else if (!strcmp(key, "PCIE_RESET_DELAY")) {
-      rshim_pcie_reset_delay = atoi(value);
-      continue;
     } else if (!strcmp(key, "USB_RESET_DELAY")) {
       rshim_usb_reset_delay = atoi(value);
       continue;
-#ifdef HAVE_RSHIM_PCIE
```
shravankumarr, line 2734: is the HAVE_RSHIM_PCIE flag not needed anymore? if so, could remove it from the Makefile also

lsun100: This HAVE_RSHIM_PCIE is still needed to compile image for BMC which doesn't have the pcie library. 
 (./configure --enable-pcie=no).  The HAVE_RSHIM_PCIE is not needed here is because I moved rshim_pcie_intr_poll_interval from rshim_pcie.c to rshim.c to put all the rshim.conf configuration variable together. This is just for code cleanup.  Another way is to put the rshim_pcie_xxx variables in rshim_pcie.c, but then we'll need to add HAVE_RSHIM_PCIE in various places in order to compile for "./configure --enable-pcie=no".

shravankumarr (approved):

PR #95, opened 2022-09-12
Adjust default boot timeout to 150s

lsun100:
The 'BOOT_TIMEOUT' configuration parameter is used to detect ARM stuck/crash that it hasn't polled the boot fifo for such time. It's reported that when redfish is enabled it could take more than 100s doing something without reading the boot stream. Adjust the default value to avoid the false error.

### DIFF
--- a/etc/rshim.conf
+++ b/etc/rshim.conf
@@ -6,7 +6,7 @@
 # Default configuration for a rshim device
 #
 #DISPLAY_LEVEL 0
-#BOOT_TIMEOUT  100
+#BOOT_TIMEOUT  150
 #DROP_MODE     0
 #USB_RESET_DELAY  3
 #PCIE_RESET_DELAY 10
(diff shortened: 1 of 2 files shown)


hselasky:
Looks good to me.

hselasky:
The time you need to wait probably depends on how many PCI functions are present? If there are two broken PCI devices timing out waiting for the firmware to load, it will take twice the time ...

lsun100:
> The time you need to wait probably depends on how many PCI functions are present? If there are two broken PCI devices timing out waiting for the firmware to load, it will take twice the time ...

Thanks for the comments. This timeout checking (bd->boot_timeout) is 'per-device'.  Each 'bd' is for separate rshim device.  My understanding is that one broken PCI device shouldn't affect the other in this case.

asmaamellanox (approved):

PR #99, opened 2022-10-03
pcie: Support mixed vfio and direct mapping mode

lsun100:
When both VFIO/UIO memory mapping and direct mapping co-exist on host with multiple rshim devices, the direct-mapped rshim will be affected when doing the VFIO bind by adding the 'vendor device' ID into sysfs new_id, which adds the direct-mapped rshim PF into VFIO as well due to 'no driver' is used.

This commit writes a faked 'rshim' string into the sysfs 'driver_override' for rshim PF to prevent other drivers from taking over it while being used.

### DIFF
--- a/src/rshim_pcie.c
+++ b/src/rshim_pcie.c
@@ -267,6 +267,15 @@ static void rshim_pcie_bind(rshim_pcie_t *dev, bool enable)
 
   if (dev->mmap_mode == RSHIM_PCIE_MMAP_VFIO ||
       dev->mmap_mode == RSHIM_PCIE_MMAP_UIO) {
+    if (!enable) {
+      snprintf(cmd, sizeof(cmd),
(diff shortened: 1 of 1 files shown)


hselasky:
This should not be enabled for FreeBSD? Which tool is used?

lsun100:
> This should not be enabled for FreeBSD? Which tool is used?

Yes, these changes are Linux-only, and are within the "#ifdef __linux__" block in the code.

lsun100:
> > This should not be enabled for FreeBSD? Which tool is used?
> 
> Yes, these changes are Linux-only, and are within the "#ifdef **linux**" block in the code.

Actually I might misunderstand the question.  Currently VFIO/UIO support is only added for Linux.  Does FreeBSD has support for them as well?  The reason to add the support is mainly due to Linux kernel lockdown feature when it has secure boot enabled. It's probably not applicable to FreeBSD.

hselasky:
I'm not sure.

PR #103, opened 2022-11-18
BF3: Support 4B access for PCIe

lsun100:
This commit adds 4B access and OPN display for BF3.

### DIFF
--- a/src/rshim.h
+++ b/src/rshim.h
@@ -184,6 +184,7 @@ enum {
 #define RSHIM_YU_BASE_ADDR             0x2800000
 #define RSHIM_YU_BOOT_RECORD_OPN       0xfd8
 #define RSHIM_YU_BOOT_RECORD_OPN_SIZE  16
+#define RSHIM_YU_BF3_BOOT_RECORD_OPN   0x9bdc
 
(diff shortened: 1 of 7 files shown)


src/rshim.h
```diff
@@ -184,6 +184,8 @@ enum {
 #define RSHIM_YU_BASE_ADDR             0x2800000
 #define RSHIM_YU_BOOT_RECORD_OPN       0xfd8
 #define RSHIM_YU_BOOT_RECORD_OPN_SIZE  16
+#define RSHIM_YU_BF3_BASE_ADDR         0x3400000
```
asmaamellanox, line 187: This is unused. Why do we need to add it?

asmaamellanox (commented):
LGTM besides small comment. Thanks Liming!

src/rshim.h
lsun100: Removed.  I was trying to clean up bf3_rshim_pcie_chan_map[] in rshim_pcie.c to use offset calculation, but don't want to put too much unrelated changes in this commit. Maybe revisit it later.

shravankumarr (approved):

PR #104, opened 2022-11-21
Add rshim log module for PSC

lsun100:
This commit adds rshim log module for PSC, so it could display logs on BF3 like "INFO[PSC]: PSC BL1 START"


asmaamellanox (approved):
so the required delay in BF1/BF2 has been fixed in BF3 hardware for both USB and PCIe?

lsun100:
> 

There are SW/FW/HW issues related to it.  SW/FW issues have been fixed. There are still a potential HW issue in BF1/BF2 and work-arounded in ATF. So far I haven't noticed such issue on BF3 yet, thus reducing the reset delay since PSC has a mailbox timeout once secure boot is enabled.

PR #107, opened 2023-03-28
Avoid opening /dev/uio multiple times

lsun100:
External tools or regression script might toogle the value of DROP_MODE in the rshim misc file, which will trigger function call rshim_pcie_mmap_uio() when UIO is used, thus reopening the /dev/uio file and causing a file handler leak. This commit adds a check to avoid such case. It does't close the handler because the interrupt thread might be still using it.

This commit also adds some comments in the configuration example rshim.conf.

### DIFF
--- a/etc/rshim.conf
+++ b/etc/rshim.conf
@@ -3,15 +3,49 @@
 #
 
 #
-# Default configuration for a rshim device
+# Display level for the 'misc' file
 #
 #DISPLAY_LEVEL 0
+
+#
+# Timeout in seconds when pushing BFB while target side is not reading the
+# boot stream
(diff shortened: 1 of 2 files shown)


hselasky:
Looks good.

One question:
Should you close() before open() in case uio_num changes?

lsun100:
> Looks good.
> 
> One question: Should you close() before open() in case uio_num changes?

Make sense. close+open would be better, but need one extra change. I'll update a new revision once the changes are verified.

hselasky (approved):
Looks good!

shravankumarr (approved):

PR #111, opened 2023-09-11
Fix interrupt handling for NIC FW_RESET

lsun100:
This commit has changes to handle the RSHIM_PCIE_RST_STATE_START event, which is needed by the new NIC FW_RESET flow which takes longer now due to the graceful shutdown action to DPU(ARM).

### DIFF
--- a/src/rshim_pcie.c
+++ b/src/rshim_pcie.c
@@ -85,7 +85,8 @@ enum {
 enum {
   RSHIM_PCIE_RST_REPLY_NONE,
   RSHIM_PCIE_RST_REPLY_ACK,
-  RSHIM_PCIE_RST_REPLY_NACK
+  RSHIM_PCIE_RST_REPLY_NACK,
+  RSHIM_PCIE_RST_START_ACK,
 };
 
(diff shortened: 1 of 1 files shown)


src/rshim_pcie.c
```diff
@@ -787,7 +787,15 @@ static void rshim_pcie_intr(rshim_pcie_t *dev)
     (info.rst_type == RSHIM_PCIE_RST_TYPE_NIC_RESET) ? "NIC" :
     ((info.rst_type == RSHIM_PCIE_RST_TYPE_DPU_RESET) ? "DPU" : ""));
 
-  if (info.rst_reply == RSHIM_PCIE_RST_REPLY_NONE) {
+  rc = bd->read_rshim(bd, RSHIM_CHANNEL, bd->regs->scratchpad6,
```
asmaamellanox, line 790: isnt this a duplicate of line 772->778?

lsun100: good catch, removed.

asmaamellanox (dismissed):

asmaamellanox (approved):

PR #117, opened 2023-11-06
Several robust fixes

lsun100:
This commit has some fixes for potential high CPU utilization.
- Add small loop delay in function rshim_fd_full_read()/ rshim_fd_full_write().
- Check checksum (either 0 or valid checksum) for Rx message.
- Check message length for Rx message.

RM #3613738


DLNvda (approved):
LGTM.

PR #123, opened 2024-01-10
Fix incorrect console message drop

lsun100:
This commit fixed a checking for oversized packet which was introduced by commit 9f19cfb4a75687ae. It should only be applied to network packet.


asmaamellanox (approved):
Thanks Liming!

PR #126, opened 2024-02-15
Update scripts/bfb-install to support remote rshim and improve speed

pgeng-nv:
* Added remote rshim so we can run BFB update remotely. Here "remote" means the RSHIM driver is running in a different processor other than the processor initiating bfb-install. Both processors will be in the same network, though they could be inside the same server enclosure. The remote rshim is invoked with an updated syntax for the "--rshim" option.

* Improved BMC update speed if using netcat instead of scp to transmit the BFB data. This is done by specifying a optional new argument of "--remote-mode" with "nc" or "ncpipe" values.

### DIFF
--- a/man/bfb-install.8
+++ b/man/bfb-install.8
@@ -1,28 +1,135 @@
 .\" Manpage for bfb-install.
-.TH man 8 "19 Nov 2020" "2.0" "bfb-install man page"
+.TH man 8 "2 Feb 2024" "3.0" "bfb-install man page"
 .SH NAME
 bfb-install \- BFB installing script for BlueField SoC over rshim driver
+
 .SH SYNOPSIS
-rshim [options]
+.B bfb-install
+-b, --bfb <bfb_file> -r, --rshim [<ip:>[:port]:]rshim<N> [-m, --remote-mode <scp|nc|ncpipe>] [-R, --reverse-nc] [-c, --config <config_file>] [-f, --rootfs <rootfs_file>] [-h, --help] [-v, --verbose]
+
+
 .SH DESCRIPTION
-bfb-install is a utility script to install BFB images on BlueField SoC over the rshim driver.
+
+bfb-install is a utility script to install BFB images on BlueField SoC over the
+rshim driver.
+
+When the "--rshim" option doesn't provide an "ip" argument, the script will run
+in local mode and try to access the local rshim device and install the BFB image
+through it.
+
+When the "--rshim" option provides an "ip" argument, the script will run in
+remote mode and try to access the remote rshim device and install the BFB image
+through it.  There are three remote modes: "scp", "nc", and "ncpipe". The
+default remote mode is "scp". "scp" uses the scp command to transfer the BFB
+image to the remote rshim device. "nc" uses the nc command to transfer the BFB
+image to the remote rshim device. "ncpipe" is similar to "nc" but uses a named
+pipe on the remote device to help transfer the BFB image and achieve faster
+speed.
+
+The "--reverse-nc" flag, used with "nc" or "ncpipe" remote modes, enables a
+reversed connection method where the local host operates a netcat server and the
+remote host, as a client, initiates the connection. This method is particularly
+useful for environments with firewall restrictions on the remote host side like
+for a BMC.
+
 .SH OPTIONS
+.TP
 -b, --bfb
-.in +4n
+
 This is the BFB image to use, which is pushed as the boot stream.
-.in
 
+.TP
 -c, --config
-.in +4n
+
 This is an optional configuration file to use, usually called bf.cfg.
-.in
 
+.TP
 -f, --rootfs
-.in +4n
-This is the optional rootfs tar.xz file which is uaually used when installing Yocto.
-.in
 
+This is the optional rootfs tar.xz file which is uaually used when installing
+Yocto.
+
+.TP
+-h, --help
+Show the help message.
+
+.TP
+-m , --remote-mode
+
+Specify the remote mode to use. The default mode is "scp". The available modes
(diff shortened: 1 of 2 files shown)


scripts/bfb-install
```diff
@@ -27,130 +27,541 @@
...
+#   $bfb, $cfg, $rootfs, $pv, $ip, $port, $rshim_node
+push_boot_stream_via_remote_rshim_nc()
+{
+  data="${bfb} ${cfg:+$cfg} ${rootfs:+${rootfs}}"
+  # Push the boot stream to remote rshim via netcat
+  echo "Pushing $data to ${ip} via netcat"
+
+  echo "Starting a remote netcat server..."
+  start_nc_server_cmd="nc -l -p $port > ${rshim_node}/boot &"
+  run_cmd_exit remote "$start_nc_server_cmd"
```
lsun100, line 180: Have you tried this remote rshim from one server to another server (instead of rshim)? line 179-180 run ssh command and exit. Just curious whether the "nc" command will terminate if the parent process (ssh) exit on a standard server. If this is the case, you might consider adding 'nohup' before the 'nc'.

scripts/bfb-install
```diff
@@ -27,130 +27,541 @@
 # of the authors and should not be interpreted as representing official policies,
 # either expressed or implied, of the FreeBSD Project.
 
+#
+# Configurations
+#
```
lsun100, line 32: A general comment:
The bfb-install script runs by /bin/sh (line 1), which might behave differently on Ubuntu server.
To be compatible, run "checkbashisms <this bfb-install script>" to check and fix the reported issues.
If there're issues and it's hard to fix, we could consider use /bin/bash instead of/bin/sh.

scripts/test-bfb-install.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
```
lsun100, line 2: Better not to add the test code into the rshim repo.

scripts/pipe_to_rshim.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+#
+# This script reads from a named pipe and writes to the rshim device.
+#
+# It is intended to be run on BMC to forward data from the host to the BMC RSHIM
+# device.
```
lsun100, line 6: Have you considered auto-generating this script at runtime with a function in bfb-install? Or else we'll need to use a better name, and add it into .deb and .rpm package which seems a little conplicated.

scripts/bfb-install
pgeng-nv: The new PR commit (67d8ab7) has been tested between Server and BMC as well between Server and Server. There were indeed some issues initially with the SSH execution not terminating as well as other issues which adding "nohup" is part of the solution.

scripts/bfb-install
pgeng-nv: I didn't know we can use bash :). I've changed it to bash. That makes everything much easier.

scripts/test-bfb-install.sh
pgeng-nv: Removed.

scripts/pipe_to_rshim.sh
pgeng-nv: Done.

pgeng-nv:
This new commit (67d8ab7) adds the capability of reverse netcat connection (using `--reverse-nc` option), as well as addressing all the PR comments.

man/bfb-install.8
```diff
@@ -1,28 +1,135 @@
...
+.RS
+bfb-install --bfb bluefield.bfb --rshim 10.15.8.200:9709:rshim0 --remote-mode ncpipe --reverse-nc
+.RE
+
+.SH KNOWN ISSUES
+
+When using remote mode nc or ncpie, there's a short window of time when there is
+a netcat server running on the remote host on the specified TCP port (default
+9527). This could be a potential security risk. So these two methods (nc or
+ncpipe) method should only be used when the remote host is on a secure network
```
lsun100, line 129: The 'method' word seems extra?

scripts/bfb-install
```diff
@@ -27,130 +27,647 @@
...
+    # must put remote nc in the background to avoid blocking the script
+    nc_client_cmd="nohup nc $local_ip $port 2>/dev/null | dd bs=1M of=${rshim_node}/boot &>/dev/null"
+    run_cmd_exit remote "$nc_client_cmd" &
+    wait_for_remote_process nc $timeout \
+      "Error: Failed to start the remote netcat server"
+  fi
+}
+
+#!/bin/bash
+
```
lsun100, line 219: These two lines should be removed

scripts/bfb-install
```diff
@@ -27,130 +27,647 @@
...
+#!/bin/sh
+#
+# This script reads from a named pipe and writes to the rshim device.
+#
+# It is intended to be run on BMC to forward data from the host to the BMC RSHIM
+# device. 
+#
+# Known Issues:
+#  - This script must be copied over to the BMC and run from there. We need to 
+#    change it to SSH-run this script from the host.
```
lsun100, line 237: Are these comments still valid (like line 228)?

man/bfb-install.8
pgeng-nv: Will fix.

scripts/bfb-install
pgeng-nv: Will fix

scripts/bfb-install
pgeng-nv: Good catch. Will fix.

lsun100 (approved):
Try to squash with the next commit for merge.

PR #131, opened 2024-04-04
DROP_MODE: sync-up the Rx FIFO when clearing DROP_MODE

lsun100:
Since DROP_MODE is a special mode to block rshim access, there could be leftover data in the Rx FIFO which causes packet out-of-sync when rshim access is re-enabled. This commit avoid such issue by draining the Rx FIFO when DROP_MODE is cleared.

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -952,14 +952,18 @@ static int rshim_fifo_tx_avail(rshim_backend_t *bd)
   return avail;
 }
 
-int rshim_fifo_sync(rshim_backend_t *bd)
+int rshim_fifo_sync(rshim_backend_t *bd, bool drop_rx)
 {
   rshim_tmfifo_msg_hdr_t hdr;
   int i, avail, rc;
+  time_t t0, t1;
+  uint64_t reg;
 
+  /* Clear pending network Rx/Tx state. */
   bd->net_rx_len = 0;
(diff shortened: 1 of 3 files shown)


src/rshim.c
```diff
@@ -975,6 +979,33 @@ int rshim_fifo_sync(rshim_backend_t *bd)
...
+        time(&t1);
+        if (difftime(t1, t0) > 1)
+          break;
+        continue;
+      }
+
+      while (avail--)
+        bd->read_rshim(bd, RSHIM_CHANNEL, bd->regs->tm_tth_data, &reg, RSHIM_REG_SIZE_8B);
+
+      time(&t0);
```
xiangrongl, line 1007: If there is no available data in the Rx FIFO, and the time difference (t1 - t0) is less than 1 second, the starting time t0 is refreshed. Does this mean: when avail = 0, the loop only exit if the time duration from the begining of the do-while loop to the time(&t1) is greater than 1 second?

xiangrongl (approved):

src/rshim.c
lsun100: Yes, the worst case is up to 2 second which is ok.  Using ">=1" condition is a little risky since the gap might be too small (like the actual t0=0.9, t1=1.1, though time() function returns seconds, unless we use a different function)

PR #134, opened 2024-04-12
Set PCIE bit in scratchpad6 and some cleanup fixes

lsun100:


src/rshim_pcie.c
```diff
@@ -794,6 +820,7 @@ static void rshim_pcie_intr(rshim_pcie_t *dev)
   switch (info.rst_state) {
   case RSHIM_PCIE_RST_STATE_REQUEST:
     RSHIM_INFO("NIC reset ACK\n");
+    info.pcie = 1;
```
pgeng-nv, line 823: `info.pcie = 1` is in all three defined cases here. If the intent is to run always, including the `default` case, then we can move it outside of the `switch`.

pgeng-nv (approved):

PR #135, opened 2024-04-12
Fix issue with incorrect IP address resolution in multi-hop routing scenarios

pgeng-nv:
This commit addresses a bug where local host's IP addresses were incorrectly resolved when there are multiple routing hops between local host and remote rshim host.

### DIFF
--- a/scripts/bfb-install
+++ b/scripts/bfb-install
@@ -139,11 +139,29 @@ run_cmd_exit()
 # Global variables used: $ip
 get_local_ip()
 {
-  local_ip=$(ip route get $ip | awk '{print $5; exit}')
-  if [ "$local_ip" == "src" ]; then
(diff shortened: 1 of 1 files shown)


lsun100 (commented):
Some small comment for the commit message
1. Suggest remove the 'scenarios' line to keep it simple.
2. Do we have Redmine for this change? If yes, better to add it here for tracking purpose.

pgeng-nv:
> Some small comment for the commit message
> 
> 1. Suggest remove the 'scenarios' line to keep it simple.
> 2. Do we have Redmine for this change? If yes, better to add it here for tracking purpose.

1. Commit message simplified.
2. No, I found it myself :)

lsun100 (approved):

PR #137, opened 2024-04-22
Prevent bmc host rshim running simultaneously

pgeng-nv:
https://redmine.mellanox.com/issues/3870080
Also, add workarounds RSHIM PF write/read bugs.

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -2155,34 +2155,28 @@ static int rshim_check_locked_mode(rshim_backend_t *bd)
 
 static int rshim_update_locked_mode(rshim_backend_t *bd)
 {
-  int rt;
-  bool locked_mode;
+  int locked_mode;
 
-  rt = rshim_check_locked_mode(bd);
-  if (rt < 0) { /* Failed to check NIC locked mode status */
+  locked_mode = rshim_check_locked_mode(bd);
+  if (locked_mode < 0)
     return -EIO;
-  }
 
-  locked_mode = !!rt;
-
-  if (locked_mode != bd->locked_mode || !bd->first_update_done) {
+  if (locked_mode != bd->locked_mode) {
     RSHIM_INFO("rshim%d set to %s mode\n", bd->index,
         locked_mode ? "locked" : "unlocked");
+    bd->locked_mode = locked_mode;
 
     /*
      * When NIC has exited locked mode, other rshim driver like BMC USB rshim
      * driver may have attached RSHIM. In that case, we will enter drop mode
      */
-    if (!locked_mode && bd->locked_mode ) {
+    if (!locked_mode) {
       if (rshim_access_check(bd)) {
         RSHIM_INFO("rshim%d attached by another device. Entering Drop Mode\n",
             bd->index);
         rshim_set_drop_mode(bd, 1);
       }
     }
-
-    bd->locked_mode = locked_mode;
-    bd->first_update_done = true;
   }
 
   return 0;
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -1704,7 +1704,7 @@ static void rshim_work_handler(rshim_backend_t *bd)
 
   if (bd->keepalive && bd->has_rshim && !bd->debug_code) {
     bd->write_rshim(bd, RSHIM_CHANNEL, bd->regs->scratchpad1,
-        RSHIM_KEEPALIVE_MAGIC_NUM, RSHIM_REG_SIZE_8B);
+                    RSHIM_KEEPALIVE_MAGIC_NUM, RSHIM_REG_SIZE_8B);
     bd->keepalive = 0;
   }
 
@@ -2431,6 +2431,7 @@ int rshim_access_check(rshim_backend_t *bd)
     rc = bd->read_rshim(bd, RSHIM_CHANNEL, bd->regs->scratchpad1, &value, RSHIM_REG_SIZE_8B);
 
     if (!rc && value == RSHIM_KEEPALIVE_MAGIC_NUM) {
+      RSHIM_INFO("another backend already attached\n");
       return -EEXIST;
     }
 
@@ -2464,12 +2465,8 @@ int rshim_register(rshim_backend_t *bd)
   }
 
   rc = rshim_access_check(bd);
-  if (rc) {
-    if (rc == -EEXIST) {
-      RSHIM_INFO("another backend already attached\n");
-    }
+  if (rc)
     return rc;
-  }
 
   if (!bd->write)
     bd->write = rshim_write_default;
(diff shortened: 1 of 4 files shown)


src/rshim.c
```diff
@@ -2359,8 +2464,12 @@ int rshim_register(rshim_backend_t *bd)
   }
 
   rc = rshim_access_check(bd);
-  if (rc)
+  if (rc) {
+    if (rc == -EEXIST) {
+      RSHIM_INFO("another backend already attached\n");
```
asmaamellanox, line 2469: why did you move this message here? I see that rshim_access_check is called in rshim_fuse.c as well

asmaamellanox (commented):
besides minor comments, LGTM

src/rshim_fuse.c
```diff
@@ -716,20 +715,11 @@ static int rshim_fuse_misc_read(struct cuse_dev *cdev, int fflags,
...
-      value = 0;
-    }
-    in_secure_nic_mode = (value == BF3_RSH_SECURE_NIC_MODE_MAGIC_NUM);
-    if (in_secure_nic_mode) {
-      n = snprintf(p, len, "%-16s%d (0:no, 1:yes)\n", "SECURE_NIC_MODE", 1);
-      p += n;
-      len -= n;
-    }
-  }
+    n = snprintf(p, len, "%-16s%d (0:no, 1:yes)\n", "SECURE_NIC_MODE",
```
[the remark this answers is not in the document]
pgeng-nv: That one probably is https://redmine.mellanox.com/issues/3871121. We have rejected that as the behavior was intentional per discussion with Liming. @Liming, do we want to keep the code here as is, or you still prefer to only show the locked status when it's in locked mode?

src/rshim.c
pgeng-nv: Good question. In the previous design, this message was printed in `rshim_access_check()`, which could cause log-overwhelming issues when we have to call it (and fail) many times. That's exactly what my previous monitoring code does. It had to call `rshim_access_check()` every 300ms when in that "zombie" mode (aka, NIC state changed from locked to unlocked, but rshim driver unable to be attach due to BMC already attached). Now the new code will enter drop mode directly, so technically this change is not needed any more. While it's still a good idea to decouple the logic and message print, I will reverse this change so we only do minimum to achieve the desired fix.

src/rshim.c
```diff
@@ -1698,7 +1704,7 @@ static void rshim_work_handler(rshim_backend_t *bd)
 
   if (bd->keepalive && bd->has_rshim && !bd->debug_code) {
     bd->write_rshim(bd, RSHIM_CHANNEL, bd->regs->scratchpad1,
-                    RSHIM_KEEPALIVE_MAGIC_NUM, RSHIM_REG_SIZE_8B);
+        RSHIM_KEEPALIVE_MAGIC_NUM, RSHIM_REG_SIZE_8B);
```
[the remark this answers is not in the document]
pgeng-nv: Addressed in new commit https://github.com/Mellanox/rshim-user-space/pull/137/commits/b74784c3aa5267b07f94c53ae61eaae18233d787.

src/rshim.c
```diff
@@ -2084,9 +2096,100 @@ rshim_backend_t *rshim_find_by_dev(void *dev)
...
+
+static int rshim_update_locked_mode(rshim_backend_t *bd)
 {
+  int rt;
+  bool locked_mode;
+
+  rt = rshim_check_locked_mode(bd);
+  if (rt < 0) { /* Failed to check NIC locked mode status */
+    return -EIO;
+  }
```
lsun100, line 2164: To be compatible with existing coding style, single line doesn't need the {}

src/rshim.c
```diff
@@ -2084,9 +2096,100 @@ rshim_backend_t *rshim_find_by_dev(void *dev)
...
+
+  locked_mode = !!rt;
+
```
lsun100, line 2167: These two lines seems unnecessary since the rshim_check_locked_mode() returns <0, 0 or 1.

src/rshim.c
```diff
@@ -2084,9 +2096,100 @@ rshim_backend_t *rshim_find_by_dev(void *dev)
...
+  if (locked_mode != bd->locked_mode || !bd->first_update_done) {
```
lsun100, line 2168: Why we need the '|| !bd->first_update_done'? Looks like it'll work well without it?

src/rshim.c
```diff
@@ -2084,9 +2096,100 @@ rshim_backend_t *rshim_find_by_dev(void *dev)
...
+    RSHIM_INFO("rshim%d set to %s mode\n", bd->index,
+        locked_mode ? "locked" : "unlocked");
+
+    /*
+     * When NIC has exited locked mode, other rshim driver like BMC USB rshim
+     * driver may have attached RSHIM. In that case, we will enter drop mode
+     */
+    if (!locked_mode && bd->locked_mode ) {
```
lsun100, line 2176: Seems that both conditions are not necessary. We only need one.

src/rshim.c
```diff
@@ -2084,9 +2096,100 @@ rshim_backend_t *rshim_find_by_dev(void *dev)
...
+      if (rshim_access_check(bd)) {
+        RSHIM_INFO("rshim%d attached by another device. Entering Drop Mode\n",
+            bd->index);
+        rshim_set_drop_mode(bd, 1);
+      }
+    }
+
+    bd->locked_mode = locked_mode;
+    bd->first_update_done = true;
```
lsun100, line 2185: this line seems not necessary

src/rshim.h
```diff
@@ -408,6 +408,10 @@ struct rshim_backend {
   /* Up to two VLAN IDs for PXE purpose. */
   uint16_t vlan[2];
 
+  /* Secure NIC mode Management. No RSHIM HW access */
+  bool locked_mode;
+  bool first_update_done;
```
lsun100, line 413: This line seems not necessary and the name of this variable causes confusion too.

src/rshim_pcie.c
```diff
@@ -861,6 +861,7 @@ static void rshim_pcie_intr(rshim_pcie_t *dev)
       delay = RSHIM_PCIE_RST_START_MIN_DELAY;
     sleep(delay);
     bd->drop_mode = drop_mode;
+    bd->locked_mode = 0;
```
lsun100, line 864: no need for this line I think. line 856~863 doesn't really change the drop_mode. It just temporarily enable drop_mode for some clearn-up and changed back.

src/rshim.c
pgeng-nv: @lsun100 , all your comments have been addressed in c73cf00

lsun100 (approved):

PR #140, opened 2024-04-25
bfb-install: Remove the monitor of 'In Enhanced NIC mode'

lsun100:
This commit removes the monitor of 'In Enhanced NIC mode' message due to behavior change that this message is not the last one anymore when installing BFB in enhanced NIC mode.


pgeng-nv (approved):

lsun100:
not needed.

PR #146, opened 2024-05-11
some robust fixes

lsun100:
This pull has two robust fixes:
1. Add a check to avoid polling the 'blocked' state during reset;
2. Adjust the default PCIe reset delay to avoid boot stream corruption.

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -661,7 +661,7 @@ int rshim_reset_control(rshim_backend_t *bd)
   int rc;
 
   rc = bd->read_rshim(bd, RSHIM_CHANNEL, bd->regs->reset_control, &reg, RSHIM_REG_SIZE_8B);
-  if (rc < 0) {
+  if (rc < 0 || RSHIM_BAD_CTRL_REG(reg)) {
     RSHIM_ERR("failed to read rshim reset control error %d\n", rc);
     return rc;
   }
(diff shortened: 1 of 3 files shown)


kblaiech (approved):
I understand this is a W/A good for the short term. Because any (future) change in FW may break this delay. FW should provide an indication as it would be a good solution for the long term.

I will leave the second commit to Penghe to review :)

src/rshim.c
```diff
@@ -2142,11 +2142,11 @@ int rshim_set_drop_mode(rshim_backend_t *bd, int value)
 static int rshim_check_locked_mode(rshim_backend_t *bd)
 {
     int rc;
-    uint64_t value;
+    uint64_t value = 0;
 
     rc = bd->read_rshim(bd, RSHIM_CHANNEL, bd->regs->scratchpad1, &value,
                         RSHIM_REG_SIZE_8B);
-    if (rc < 0) {
+    if (rc < 0 || RSHIM_BAD_CTRL_REG(value)) {
```
pgeng-nv, line 2149: It's not clear why "RSHIM_BAD_CTRL_REG" is specifically added here but not in other "bd->read_rshim()" calls. If there's a specific reason we only need it here in "rshim_check_locked_mode()", maybe explain it here. Or maybe move this check inside rshim_pcie_read()?

pgeng-nv (commented):
Need comment on "RSHIM_BAD_CTRL_REG".

src/rshim.c
lsun100: RSHIM_BAD_CTRL_REG is used in several other places as well, which could potentially happen in certain corner cases such as during ARM reset. It's mainly used for registers that we know the content. It's not reliable to be used for generic register read.

pgeng-nv (dismissed):

pgeng-nv (approved):

PR #147, opened 2024-05-16
pcie-lf driver enhancement to support register read/write with the bf_reg.py script

lsun100:
This one is needed by RM #3892195 to support register read/write by pcie-lf driver. 

During the testing for livefish driver, two more fixes are also made:
1. Cleanup the code to remove the workaround for BF2 A0;
2. Add a check to do 'blocked' state checking only for PCIe backend.

### DIFF
--- a/src/rshim_pcie_lf.c
+++ b/src/rshim_pcie_lf.c
@@ -759,6 +759,7 @@ static int rshim_pcie_probe(struct pci_dev *pci_dev)
     bd = &dev->bd;
     strcpy(bd->dev_name, dev_name);
     /* Enable drop mode by default if not configured. */
+    bd->type = RSH_BACKEND_PCIE_LF;
     bd->drop_mode = (rshim_drop_mode >= 0) ? rshim_drop_mode : 1;
     bd->read_rshim = rshim_pcie_read;
     bd->write_rshim = rshim_pcie_write;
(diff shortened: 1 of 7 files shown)


src/rshim_pcie_lf.c
```diff
@@ -186,6 +186,8 @@ static int msn_gw_read(struct pci_dev *pci_dev, int addr,
   uint32_t data;
   int rc = 0;
 
+  addr &= ~BF3_RSH_ADDR_MASK;
```
pgeng-nv, line 189: This bit-clear of an address looks having a specific purpose. Better to add a comment here.

src/rshim_pcie_lf.c
```diff
@@ -644,6 +648,19 @@ static int rshim_boot_fifo_write(struct pci_dev *pci_dev, int addr,
   return 0;
 }
 
+static uint32_t
```
pgeng-nv, line 651: Would love to see a function comment here.

pgeng-nv (approved):
+1

src/rshim_pcie_lf.c
lsun100: done

src/rshim_pcie_lf.c
lsun100: done

shravankumarr (approved):

pgeng-nv (approved):

PR #149, opened 2024-06-06
bfb-install: add runtime image support

lsun100:
This commit enables runtime image upgrade with the '-u' option, so image can be pushed without resetting the DPU. It'll expect rshim log 'Runtime upgrade finished' to exit the polling.

A utility script bf_reg.pl is added to access registers via rshim. It's in perl, so it could be used on DPU BMC as well by default.

It also has a new misc command 'CLEAR_ON_READ'. If set, the rshim log will be cleared after read.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,7 +7,7 @@ SUBDIRS = src man
 dist_doc_DATA = README.md
 confdir = $(sysconfdir)
 dist_conf_DATA = etc/rshim.conf
-dist_sbin_SCRIPTS = scripts/bfb-install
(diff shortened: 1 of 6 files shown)


pgeng-nv (dismissed):
+1.  Looks good to me. Better mention the 2nd feature of "CLEAR_ON_READ" in commit message, or split the commit to two separate ones.

lsun100:
> +1. Looks good to me. Better mention the 2nd feature of "CLEAR_ON_READ" in commit message, or split the commit to two separate ones.

Sure, I'll remove the 'CLEAR_ON_READ' and make it separately.

PR #151, opened 2024-06-08
bfb-install: add cleanup code for runtime update

lsun100:
This commit adds cleanup code for runtime update, so SP2.BIT2 could be always cleared when script completes.

### DIFF
--- a/scripts/bfb-install
+++ b/scripts/bfb-install
@@ -181,9 +181,6 @@ clear_boot_fifo()
 }
 
 # Push the boot stream to rshim via local rshim
-#
-# Global variables used:
-#   $bfb, $cfg, $rootfs, $pv, $rshim_node, $sudo_prefix
 push_boot_stream_via_local_rshim()
 {
   # Push the boot stream to local rshim
@@ -196,6 +193,8 @@ push_boot_stream_via_local_rshim()
     clear_boot_fifo
     # Set SP2.BIT2=1
     ${BF_REG} $(basename ${rshim_node}) 0x13000c48.64 0x4 >/dev/null
+    # Set SWINT2.BIT2
+    ${BF_REG} $(basename ${rshim_node}) 0x13000318.64 0x4 >/dev/null
   fi
 
   $sudo_prefix sh -c "cat ${bfb} ${cfg:+$cfg} ${rootfs:+${rootfs}} ${pv:+| ${pv} | cat -} > ${rshim_node}/boot"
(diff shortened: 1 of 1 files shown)


scripts/bfb-install
```diff
@@ -196,6 +196,8 @@ push_boot_stream_via_local_rshim()
     clear_boot_fifo
     # Set SP2.BIT2=1
     ${BF_REG} $(basename ${rshim_node}) 0x13000c48.64 0x4 >/dev/null
+    # Set SWINT2.BIT2
+    ${BF_REG} $(basename ${rshim_node}) 0x13000318.64 0x4 >/dev/null
```
pgeng-nv, line 200: We may want to add BF_REG to the global variable list line above:

    # Global variables used:
    #   $bfb, $cfg, $rootfs, $pv, $rshim_node, $sudo_prefix

However, I tend to agree it's a bit of work to update this list on an ongoing basis, and its usefulness may be limited. So we could remove just remove these two comment lines instead.

scripts/bfb-install
```diff
@@ -459,6 +461,17 @@ cleanup() {
...
+  if [ $runtime -eq 1 ]; then
+    # Reset to default state.
+    echo "BOOT_RESET_SKIP 0" > ${rshim_node}/misc
+
+    # Cleanup SP2.BIT2 and trigger SWINT2.
+    sp2=`${BF_REG} $(basename ${rshim_node}) 0x13000c48.64 | awk '{print $3}'`
+    sp2=$((sp & ~4))
+    ${BF_REG} $(basename ${rshim_node}) 0x13000c48.64 $sp2 >/dev/null
+    ${BF_REG} $(basename ${rshim_node}) 0x13000318.64 0x4 >/dev/null
+  fi
```
pgeng-nv, line 474: Same comment as the first one. It applies to the newly added variable $runtime, $BF_REG etc. We can remove the two function comment lines.

scripts/bfb-install
```diff
@@ -459,6 +461,17 @@ cleanup() {
...
     fi
   fi
+
+  if [ $runtime -eq 1 ]; then
+    # Reset to default state.
+    echo "BOOT_RESET_SKIP 0" > ${rshim_node}/misc
+
+    # Cleanup SP2.BIT2 and trigger SWINT2.
+    sp2=`${BF_REG} $(basename ${rshim_node}) 0x13000c48.64 | awk '{print $3}'`
+    sp2=$((sp & ~4))
```
pgeng-nv, line 471: The `sp` in line 471 should be `sp2`?

lsun100: good catch, thanks, fixed.

scripts/bfb-install
lsun100: done

scripts/bfb-install
lsun100: done

pgeng-nv (approved):

PR #153, opened 2024-06-11
bfb-install: Enable CLEAR_ON_READ

lsun100:
rshim log buffer is only 1KB in size. Once full, it'll discard the rest of messages which might contain important messages for errors or installation completion. This commit enables CLEAR_ON_READ as a quick workaround to clear the log buffer after read during the polling of installation status. Since the log is cleared from the log buffer during bfb-install, file /tmp/bfb-install.log is create to preserve these logs in case it might be needed.

In next release, it's preferred to update the kernel driver on DPU side, such as by adding a new local buffer for  Linux logging, or making the rshim log as circular buffer for logs from Linux.

RM #3908498

### DIFF
--- a/scripts/bfb-install
+++ b/scripts/bfb-install
@@ -32,8 +32,9 @@
 #
 
 DEBUG=${DEBUG:-0}                             # Debug mode
-RSHIM_PIPE="/tmp/rshim_pipe"
 BF_REG=$(dirname $0)/bf-reg
+RSHIM_PIPE="/tmp/rshim_pipe"
+LOG_FILE="/tmp/bfb-install.log"
 
 usage ()
 {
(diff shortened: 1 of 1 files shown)


pgeng-nv (approved):
+2. Looks good to me.

Only comment is it might be helpful to indicate "CLEAR_ON_READ" is on in "misc" output, as it may confuse the user when Log Messages shows empty when the user cats the misc file during bfb-install.

lsun100:
> +2. Looks good to me.
> 
> Only comment is it might be helpful to indicate "CLEAR_ON_READ" is on in "misc" output, as it may confuse the user when Log Messages shows empty when the user cats the misc file during bfb-install.

Sorry, missed this comment. Yes, CLEAR_ON_READ will be displayed, but in "DISPLAY_LEVEL 1". Maybe I should move it to DISPLAY_LEVEL 2.

PR #155, opened 2024-06-27
Add ownership transfer feature

pgeng-nv:
When Rshim driver finds another backend has attached to the rshim device, it
will enter drop mode. In this mode, user can echo a "FORCE_CMD 1" to
the misc file to request rshim ownership from the other backend.

Similarly, user can also send a one-time Force command using command option of
"--force", or setting "FORCE_CMD" to 1 in rshim.conf.

If Rshim driver is able to attach to the rshim device, it will monitor
the SP1 for incoming "RequestOwnership" command, and release rshim when
that command is found.

RM: 3679848

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -30,16 +30,12 @@
 #define RSHIM_TIMER_INTERVAL 1
 
 /* Intervals to check the locked mode. */
-#define CHECK_LOCKED_MODE_MS      100
-#define CHECK_LOCKED_MODE_TICKS   (CHECK_LOCKED_MODE_MS / RSHIM_TIMER_INTERVAL)
+#define RSHIM_CHECK_LOCKED_MODE_MS      100
+#define RSHIM_CHECK_LOCKED_MODE_TICKS   (RSHIM_CHECK_LOCKED_MODE_MS / RSHIM_TIMER_INTERVAL)
 
-/* Timeouts in ms */
-#define TO_RSHIM_REQ_MS 5000
-#define TO_RSHIM_ONLINE_MS 5000
-
-/* Timeouts in ticks */
-#define TO_TICKS_RSHIM_REQ (TO_RSHIM_REQ_MS / OSP_MGT_INTERVAL_MS)
-#define TO_TICKS_RSHIM_ONLINE (TO_RSHIM_ONLINE_MS / OSP_MGT_INTERVAL_MS)
+/* Ownership protocol timeouts in ms */
+#define RSHIM_OSP_TO_REQ_MS 1000
+#define RSHIM_OSP_TO_ONLINE_MS 1000
 
 /* Cycles to poll the network initialization before timeout. */
 #define RSHIM_NET_INIT_DELAY (60000 / RSHIM_TIMER_INTERVAL)
@@ -55,22 +51,18 @@ static int rshim_keepalive_ticks = RSHIM_KEEPALIVE_PERIOD / RSHIM_TIMER_INTERVAL
 #define RSHIM_KEEPALIVE_MAGIC_NUM 0x5089836482ULL
 
 /* Rshim ownership management. */
-#define OSP_MGT_INTERVAL_MS        100 /* Ownership state machine interval */
-#define OSP_MGT_INTERVAL_TICKS     (OSP_MGT_INTERVAL_MS / RSHIM_TIMER_INTERVAL)
-#define RSHIM_OWNERSHIP_REQ_MAGIC_NUM 0x4F53505F524551ULL /* OSP_REQ */
-#define RSHIM_OWNERSHIP_ACK_MAGIC_NUM 0x4F53505F41434BULL /* OSP_ACK */
-
-/* Pause RSHIM_DBG() here after 10 messages, then resume after 1 minute */
-#define RSHIM_DBG_PAUSE 6   /* messages limit for burst print */
-#define RSHIM_DBG_RESUME 10   /* seconds before resuming */
+#define RSHIM_OSP_MGT_INTERVAL_MS        100 /* Ownership handler running interval */
+#define RSHIM_OSP_MGT_INTERVAL_TICKS     (RSHIM_OSP_MGT_INTERVAL_MS / RSHIM_TIMER_INTERVAL)
+#define RSHIM_OSP_REQ_MAGIC_NUM 0x4F53505F524551ULL /* OSP_REQ */
+#define RSHIM_OSP_ACK_MAGIC_NUM 0x4F53505F41434BULL /* OSP_ACK */
 
 const char* magic_to_str(uint64_t magic){
   switch(magic){
     case RSHIM_KEEPALIVE_MAGIC_NUM:
       return "MAGIC_KEEPALIVE";
-    case RSHIM_OWNERSHIP_REQ_MAGIC_NUM:
+    case RSHIM_OSP_REQ_MAGIC_NUM:
       return "MAGIC_OSP_REQ";
-    case RSHIM_OWNERSHIP_ACK_MAGIC_NUM:
+    case RSHIM_OSP_ACK_MAGIC_NUM:
       return "MAGIC_OSP_ACK";
     default:
       return "UNKNOWN";
@@ -252,7 +244,7 @@ static void rshim_fifo_msg_update_checksum(rshim_tmfifo_msg_hdr_t *hdr);
 
 static int rshim_update_locked_mode(rshim_backend_t *bd);
 
-static int handle_ownership_transfer(rshim_backend_t *bd);
+static int rshim_handle_ownership_transfer(rshim_backend_t *bd);
 
 /* Global lock / unlock. */
 void rshim_lock(void)
@@ -1791,13 +1783,13 @@ static void rshim_work_handler(rshim_backend_t *bd)
     bd->has_locked_work = 0;
     rshim_update_locked_mode(bd);
   }
-  pthread_mutex_unlock(&bd->mutex);
 
-  /* Ownership transfer needs to run without holding bd->mutex. */
   if (bd->has_osp_work) {
     bd->has_osp_work = 0;
-    handle_ownership_transfer(bd);
+    rshim_handle_ownership_transfer(bd);
   }
+
+  pthread_mutex_unlock(&bd->mutex);
 }
 
 static int rshim_boot_done(rshim_backend_t *bd)
@@ -2122,9 +2114,6 @@ rshim_backend_t *rshim_find_by_dev(void *dev)
 
 /*
  * Must run with bd->mutex held.
- *
- * No need to call this function when value is set to 1. Just set bd->drop_mode
- * to 1.
  * */
 int rshim_set_drop_mode(rshim_backend_t *bd, int value)
 {
@@ -2137,7 +2126,10 @@ int rshim_set_drop_mode(rshim_backend_t *bd, int value)
     return -EALREADY;
   }
 
-  bd->drop_mode = value;
+  if (bd->enable_device && bd->enable_device(bd, value ? false : true))
+    bd->drop_mode = 1;
+  else
+    bd->drop_mode = value;
 
   if (bd->drop_mode)
     bd->drop_pkt = 1;
@@ -2218,27 +2210,14 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
   return 0;
 }
 
-static int read_rshim_sp1(rshim_backend_t *bd, uint64_t* sp1)
-{
-  int rc;
-
-  rc = bd->read_rshim(bd, RSHIM_CHANNEL, bd->regs->scratchpad1, sp1, RSHIM_REG_SIZE_8B);
-  if (rc) {
-    RSHIM_ERR("rshim%d failed to read sp1\n", bd->index);
-    return rc;
-  }
-
-  return 0;
-}
-
-static int check_rshim_sp1_magic(rshim_backend_t *bd, uint64_t magic,
+static int rshim_check_sp1_magic(rshim_backend_t *bd, uint64_t magic,
     bool* result)
 {
   int rc;
   uint64_t sp1;
 
-  rc = read_rshim_sp1(bd, &sp1);
-  if (rc) {
+  rc = bd->read_rshim(bd, RSHIM_CHANNEL, bd->regs->scratchpad1, &sp1, RSHIM_REG_SIZE_8B);
+  if (rc || RSHIM_BAD_CTRL_REG(sp1)) {
     RSHIM_ERR("rshim%d failed to read sp1: %d\n", bd->index, rc);
     *result = false;
     return rc;
@@ -2249,19 +2228,12 @@ static int check_rshim_sp1_magic(rshim_backend_t *bd, uint64_t magic,
   return 0;
 }
 
-static int check_rshim_magic_ownership_req(rshim_backend_t *bd, bool* result)
-{
-  return check_rshim_sp1_magic(bd, RSHIM_OWNERSHIP_REQ_MAGIC_NUM, result);
-}
-
-static int check_rshim_magic_ownership_ack(rshim_backend_t *bd, bool* result)
-{
-  return check_rshim_sp1_magic(bd, RSHIM_OWNERSHIP_ACK_MAGIC_NUM, result);
-}
-
-static int write_rshim_sp1_magic(rshim_backend_t *bd, uint64_t magic)
+static int rshim_write_sp1_magic(rshim_backend_t *bd, uint64_t magic)
 {
   int rc;
+
+  RSHIM_INFO("rshim%d writing sp1 magic %s (0x%llx)\n", bd->index,
+      magic_to_str(magic), (unsigned long long)magic);
 
   rc = bd->write_rshim(bd, RSHIM_CHANNEL, bd->regs->scratchpad1, magic,
       RSHIM_REG_SIZE_8B);
@@ -2273,23 +2245,8 @@ static int write_rshim_sp1_magic(rshim_backend_t *bd, uint64_t magic)
   return 0;
 }
 
-static int write_rshim_magic_ownership_req(rshim_backend_t *bd)
+static int rshim_handle_ownership_transfer(rshim_backend_t *bd)
 {
-  RSHIM_INFO("rshim%d writing ownership req magic: 0x%llx\n", bd->index,
-      RSHIM_OWNERSHIP_REQ_MAGIC_NUM);
-
-  return write_rshim_sp1_magic(bd, RSHIM_OWNERSHIP_REQ_MAGIC_NUM);
-}
-
-static int write_rshim_magic_ownership_ack(rshim_backend_t *bd)
-{
-  RSHIM_INFO("rshim%d writing ownership ack magic 0x%llx\n", bd->index,
-      RSHIM_OWNERSHIP_ACK_MAGIC_NUM);
-
-  return write_rshim_sp1_magic(bd, RSHIM_OWNERSHIP_ACK_MAGIC_NUM);
-}
-
-static int handle_ownership_transfer(rshim_backend_t *bd) {
   int i, rt;
   bool has_ack;
   bool has_req;
@@ -2299,21 +2256,26 @@ static int handle_ownership_transfer(rshim_backend_t *bd) {
       RSHIM_INFO("rshim%d executing Force command\n", bd->index);
       bd->force_cmd_pending = 0;
 
+      if (bd->enable_device && bd->enable_device(bd, true)) {
+        RSHIM_ERR("rshim%d failed to enable device\n", bd->index);
+        return -EIO;
+      }
+
       bd->requesting_rshim = 1;
 
       /* sending req and checking ack multiple times to ensure the transfer */
       for (i = 0; i < 10; i++) {
-        rt = write_rshim_magic_ownership_req(bd);
+        rt = rshim_write_sp1_magic(bd, RSHIM_OSP_REQ_MAGIC_NUM);
         if (rt) {
           RSHIM_ERR("rshim%d failed to write ownership transfer req\n",
               bd->index);
           continue;
         }
 
-        usleep(100000);
+        usleep(RSHIM_OSP_TO_REQ_MS * 1000 / 10);
 
         has_ack = false;
-        rt = check_rshim_magic_ownership_ack(bd, &has_ack);
+        rt = rshim_check_sp1_magic(bd, RSHIM_OSP_ACK_MAGIC_NUM, &has_ack);
 
         if (!rt && has_ack) {
           break;
@@ -2339,7 +2301,7 @@ static int handle_ownership_transfer(rshim_backend_t *bd) {
     }
   } else {
     has_req = false;
-    rt = check_rshim_magic_ownership_req(bd, &has_req);
+    rt = rshim_check_sp1_magic(bd, RSHIM_OSP_REQ_MAGIC_NUM, &has_req);
 
     if (!rt && has_req) {
       RSHIM_INFO("rshim%d received ownership transfer request\n", bd->index);
@@ -2348,12 +2310,11 @@ static int handle_ownership_transfer(rshim_backend_t *bd) {
 
      RSHIM_INFO("Notifying the requester with ACK\n");
       for (i = 0; i < 10; i++) {
-        write_rshim_magic_ownership_ack(bd);
-        usleep(100000);
+        rshim_write_sp1_magic(bd, RSHIM_OSP_ACK_MAGIC_NUM);
+        usleep(RSHIM_OSP_TO_REQ_MS * 1000 / 10);
       }
 
-      /* Make sure the requester is stable */
-      usleep(1000000);
+      usleep(RSHIM_OSP_TO_ONLINE_MS * 1000);
 
       RSHIM_INFO("rshim%d entering drop mode\n", bd->index);
       rt = rshim_set_drop_mode(bd, 1);
@@ -2365,7 +2326,7 @@ static int handle_ownership_transfer(rshim_backend_t *bd) {
 
       rshim_unlock();
 
-      RSHIM_INFO("rshim%d passed to the requester\n", bd->index);
+      RSHIM_INFO("rshim%d passed to the requester successfully\n", bd->index);
     }
   }
   return 0;
@@ -2408,12 +2369,12 @@ static void rshim_timer_run(void)
       if (rshim_timer_ticks - bd->timer > 0)
         rshim_timer_func(bd);
 
-      if (rshim_timer_ticks % CHECK_LOCKED_MODE_TICKS == 0) {
+      if (rshim_timer_ticks % RSHIM_CHECK_LOCKED_MODE_TICKS == 0) {
         bd->has_locked_work = 1;
         rshim_work_signal(bd);
       }
 
-      if (rshim_timer_ticks % OSP_MGT_INTERVAL_TICKS == 0) {
+      if (rshim_timer_ticks % RSHIM_OSP_MGT_INTERVAL_TICKS == 0) {
         bd->has_osp_work = 1;
         rshim_work_signal(bd);
       }
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -1710,7 +1710,8 @@ static void rshim_work_handler(rshim_backend_t *bd)
 
   bd->work_pending = false;
 
-  if (bd->keepalive && bd->has_rshim && !bd->debug_code) {
+  if (bd->keepalive && bd->has_rshim && !bd->debug_code && !bd->drop_mode &&
+      !bd->in_access_check) {
     bd->write_rshim(bd, RSHIM_CHANNEL, bd->regs->scratchpad1,
                     RSHIM_KEEPALIVE_MAGIC_NUM, RSHIM_REG_SIZE_8B);
     bd->keepalive = 0;
@@ -2465,6 +2466,8 @@ int rshim_access_check(rshim_backend_t *bd)
   uint64_t value = 0;
   int i, rc;
 
+  bd->in_access_check = 1;  /* must be cleared before return */
+
   /*
    * Add a check and delay to make sure rshim is ready.
    * It's mainly used in BlueField-2+ where the rshim (like USB) access is
@@ -2479,6 +2482,7 @@ int rshim_access_check(rshim_backend_t *bd)
   }
   if (RSHIM_BAD_CTRL_REG(value)) {
     RSHIM_ERR("Unable to read from rshim\n");
+    bd->in_access_check = 0;
     return -ETIMEDOUT;
   }
 
@@ -2488,6 +2492,7 @@ int rshim_access_check(rshim_backend_t *bd)
   rc = bd->write_rshim(bd, RSHIM_CHANNEL, bd->regs->scratchpad1, 0, RSHIM_REG_SIZE_8B);
   if (rc < 0) {
     RSHIM_ERR("failed to write rshim rc=%d\n", rc);
+    bd->in_access_check = 0;
     return -ENODEV;
   }
 
@@ -2513,6 +2518,7 @@ int rshim_access_check(rshim_backend_t *bd)
 
     if (!rc && value == RSHIM_KEEPALIVE_MAGIC_NUM) {
       RSHIM_INFO("another backend already attached\n");
+      bd->in_access_check = 0;
       return -EEXIST;
     }
 
@@ -2523,8 +2529,11 @@ int rshim_access_check(rshim_backend_t *bd)
   rc = bd->read_rshim(bd, RSHIM_CHANNEL, bd->regs->scratchpad1, &value, RSHIM_REG_SIZE_8B);
   if (rc < 0 || RSHIM_BAD_CTRL_REG(value)) {
     RSHIM_ERR("access_check: failed to read rshim\n");
+    bd->in_access_check = 0;
     return -ENODEV;
   }
+
+  bd->in_access_check = 0;
 
   return 0;
 }
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -1304,11 +1304,10 @@ again:
    */
   if (bd->read_buf_next < bd->read_buf_bytes ||
       (bd->spin_flags & RSH_SFLG_READING)) {
-
     /* We're doing nothing. */
     RSHIM_DBG("fifo_input: no new read: %s\n",
-        (bd->read_buf_next < bd->read_buf_bytes) ?
-        "have data" : "already reading");
+              (bd->read_buf_next < bd->read_buf_bytes) ?
+              "have data" : "already reading");
   } else {
     int len;
 
@@ -1351,8 +1350,7 @@ ssize_t rshim_fifo_read(rshim_backend_t *bd, char *buffer, size_t count,
     int pass1;
     int pass2;
 
-    /* Remove this for now as it's every chatty with -l 4*/
-    /* RSHIM_DBG("fifo_read, top of loop, remaining count %zd\n", count); */
+    RSHIM_DBG("fifo_read, top of loop, remaining count %zd\n", count);
 
     /*
      * We check this each time through the loop since the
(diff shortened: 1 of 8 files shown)


lsun100 (commented):
Can we summarize the changes in the commit message?

src/rshim.h
```diff
@@ -495,7 +512,7 @@ extern volatile bool rshim_run;
 
 /* Register/unregister backend. */
 int rshim_register(rshim_backend_t *bd);
-void rshim_deregister(rshim_backend_t *bd);
+void rshim_deregister(rshim_backend_t *bd, bool include_misc);
```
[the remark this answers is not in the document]
pgeng-nv: fixed in 2c673bc

src/rshim.h
```diff
@@ -617,7 +634,7 @@ static inline void rshim_pcie_check(rshim_backend_t *bd)
 
 #ifdef HAVE_RSHIM_FUSE
 int rshim_fuse_init(rshim_backend_t *bd);
-int rshim_fuse_del(rshim_backend_t *bd);
+int rshim_fuse_del(rshim_backend_t *bd, bool include_misc);
```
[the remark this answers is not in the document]
pgeng-nv: fixed in 2c673bc.

src/rshim_fuse.c
```diff
@@ -942,6 +958,82 @@ static int rshim_fuse_misc_write(struct cuse_dev *cdev, int fflags,
...
+      goto invalid;
+    if (value) {
+      pthread_mutex_lock(&bd->mutex);
+      if (rshim_access_check(bd)) {
+        RSHIM_INFO("rshim%d attached by another device\n", bd->index);
+      } else {
+        RSHIM_INFO("rshim%d is free to use\n", bd->index);
+      }
+      pthread_mutex_unlock(&bd->mutex);
+    }
```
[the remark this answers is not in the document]
pgeng-nv: As discussed, I have removed it in c315936.

src/rshim_fuse.c
```diff
@@ -942,6 +958,82 @@ static int rshim_fuse_misc_write(struct cuse_dev *cdev, int fflags,
...
+        goto invalid;
+      }
+      if (!bd->access_check_failed) {
+        rt = rshim_notify(bd, RSH_EVENT_ATTACH, 0);
+        if (rt) {
+          RSHIM_ERR("Failed to attach rshim%d\n", bd->index);
+          goto invalid;
+        }
+      }
+    }
```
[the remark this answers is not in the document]
pgeng-nv: As discussed, I have removed these in c315936.

src/rshim.h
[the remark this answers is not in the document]
pgeng-nv: Just realized "has_misc" means the opposite of "include_misc". For rshim_fuse_del(..., bool include_misc) will delete "misc" if "include_misc" is true, but it should keep "misc" if "has_misc" is true. I have to update this.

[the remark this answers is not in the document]
pgeng-nv: Updated in f87dc1b

src/rshim_fuse.c
```diff
@@ -1354,11 +1452,13 @@ int rshim_fuse_init(rshim_backend_t *bd)
   return 0;
 }
 
-int rshim_fuse_del(rshim_backend_t *bd)
+int rshim_fuse_del(rshim_backend_t *bd, bool include_misc)
```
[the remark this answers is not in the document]
pgeng-nv: Added comment in f87dc1b.

pgeng-nv:
> Can we summarize the changes in the commit message?

Sure. I will edit the main commit message. But that will alter the commit history and my commit references in my comments above will be invalid. I will just squash them together and force push it.

pgeng-nv:
> Can we summarize the changes in the commit message?

Done in forced commit 6284de2

src/rshim_fuse.c
```diff
@@ -680,7 +680,17 @@ static int rshim_fuse_misc_read(struct cuse_dev *cdev, int fflags,
   p += n;
   len -= n;
 
-  /* SW reset flag is always 0. */
+  n = snprintf(p, len, "%-16s%d (0:no access, 1:has access)\n", "RSHIM ACCESS",
```
lsun100, line 683: The misc file os key/value style where the first string is keyword. Can we change it to be "RSHIM_ACCESS" instead of the space separated "RSHIM ACCESS"?

src/rshim_fuse.c
```diff
@@ -721,6 +731,11 @@ static int rshim_fuse_misc_read(struct cuse_dev *cdev, int fflags,
     len -= n;
   }
 
+  n = snprintf(p, len, "%-16s%d (0:no, 1:yes)\n", "FORCE_CMD",
+      bd->force_cmd_pending);
```
lsun100, line 735: A question, will the force_cmd_pending be cleared once it succeed, or later taken over again by the other side?

src/rshim_fuse.c
```diff
@@ -1305,6 +1328,10 @@ int rshim_fuse_init(rshim_backend_t *bd)
      * device was re-ceated during SW_RESET.
      */
     snprintf(buf, sizeof(buf), "/dev/rshim%d/%s", bd->index, name);
+    if (i == RSH_DEV_TYPE_MISC && !access(buf, F_OK)) {
+      /* It's OK for misc file to be persistent after fuse_del() */
+      continue;
+    }
```
lsun100, line 1334: Any reason we need this logic? It might cause issue for next restart. The fuse file will cause trouble if no rshim driver supports it.

src/rshim_fuse.c
```diff
@@ -1354,11 +1381,19 @@ int rshim_fuse_init(rshim_backend_t *bd)
   return 0;
 }
 
-int rshim_fuse_del(rshim_backend_t *bd)
+/*
+ * If has_misc is true, we will not delete the "misc" device file even if
+ * other device files like "boot" etc are deregistered. This allows user
+ * to send commands to the driver even if it's not fully attached to rshim
+ * device.
+ */
```
lsun100, line 1389: How do we handle the rshim service stop and rshim restart?
Is it possible to remove the misc device file when rshim driver is gone?

lsun100 (commented):
In addition to the existing changes, I don't see the changes related to the ownership transfer itself.  Is it implemented in this commit?

src/rshim_fuse.c
pgeng-nv: Will change.

src/rshim_fuse.c
pgeng-nv: It'll be cleared once it's executed, regardless of whether it's successful or not.

pgeng-nv:
> In addition to the existing changes, I don't see the changes related to the ownership transfer itself. Is it implemented in this commit?

Yes it's in this commit. Did you try to open the changes to "rshim.c" manually? By default, the changes pane for rshim.c is folded as too many changes are in there.

src/rshim_fuse.c
pgeng-nv: After the driver releases rshim upon other side's request, it still needs to keep the "misc" device node so the user can cat "FORCE_CMD" to it. I agree this needs to be thoroughly tested.

src/rshim_fuse.c
pgeng-nv: Yes, it should. When stopping the rshim service, `rshim_stop()` will be called, which should call `rshim_deregister(bd, false)` , where the second argument "has_misc" is different from other normal calls.

BTW, I just found I need to make additional changes for renaming "include_misc" to "has_misc". All logics have to be reversed, aka, true has to be changed to false and vice versa. I didn't do that in my previous force update.

src/rshim_fuse.c
pgeng-nv: fixed in 269e388

src/rshim.c
```diff
@@ -1290,7 +1351,8 @@ ssize_t rshim_fifo_read(rshim_backend_t *bd, char *buffer, size_t count,
     int pass1;
     int pass2;
 
-    RSHIM_DBG("fifo_read, top of loop, remaining count %zd\n", count);
+    /* Remove this for now as it's every chatty with -l 4*/
```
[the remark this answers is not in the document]
pgeng-nv: Good catch. I tried to remove those chatty message in "- 4" after discussion with Liming. But those changes will be too scattered and better to be done in another time. I undid most of those changes but somehow skipped this one. Will revert it back.

src/rshim_fuse.c
```diff
@@ -680,7 +680,17 @@ static int rshim_fuse_misc_read(struct cuse_dev *cdev, int fflags,
   p += n;
   len -= n;
 
-  /* SW reset flag is always 0. */
+  n = snprintf(p, len, "%-16s%d (0:no access, 1:has access)\n", "RSHIM ACCESS",
+               !bd->access_check_failed);
+  p += n;
+  len -= n;
+
+  n = snprintf(p, len, "%-16s%d (0:no, 1:yes)\n", "MONITOR_MODE",
```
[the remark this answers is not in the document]
pgeng-nv: Correct.

src/rshim_fuse.c
```diff
@@ -1294,6 +1315,8 @@ int rshim_fuse_init(rshim_backend_t *bd)
 #endif
 
   for (i = 0; i < RSH_DEV_TYPES; i++) {
+    if (bd->access_check_failed && i != RSH_DEV_TYPE_MISC)
```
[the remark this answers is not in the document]
pgeng-nv: Will do.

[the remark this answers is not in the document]
pgeng-nv: Addressed in f4c5742

src/rshim.c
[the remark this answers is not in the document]
pgeng-nv: addressed in f4c5742

src/rshim.c
```diff
@@ -2697,13 +2992,10 @@ static void rshim_main(int argc, char *argv[])
...
       }
     } else {
-      /* Disable the timer if no rshim devices are found. */
-      if (rshim_dev_bitmask) {
-        if (!rshim_timer_interval)
-          rshim_set_timer(timer_fd, RSHIM_TIMER_INTERVAL);
-      } else if (rshim_timer_interval) {
-          rshim_set_timer(timer_fd, 0);
-      }
+      /* Always enable the timer even if no rshim devices are found, so we can
```
asmaamellanox, line 2993: could you please explain how this works. My understanding from the code:
1) Rshim USB on BMC is enabled
2) user enables rshim pcie on x86  host and uses FORCE CMD to take control over rshim.
3) The ownership transfer is done through writes of scratchpad1.
4) Now the user has no access from BMC rshim USB except for dumping misc. misc shows that it is in monitor mode (=1) and rshim access = 0. For rshim pcie, monitor mode =0 , force mode =1, rshim access = 1.

what matters to me is what happens after this:
1) the user has the ability to take back the BMC rshim USB by setting FORCE_CMD=0? 
2) and this is why we keep the misc file alive even if the rshim device is deregistered. So the only file we would see under /dev/rshimX/ is misc?

my question is why dont we just transfer the ownership completely to rshim pcie once we set FORCE_CMD=1? and deregister properly the rshim USB. if the user needs to take control again from rshim usb, they can just issue FORCE_CMD=1 from the rshim usb itself.

src/rshim.c
```diff
@@ -2138,9 +2211,7 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
   if (bd->is_booting)
     return 0;
 
-  pthread_mutex_lock(&bd->mutex);
```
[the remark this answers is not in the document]
pgeng-nv: The lock is still there. The code has been refactored. Previously this function is called directly in the timer function. Now it's called in the worker handler, which is already protected in the same bd->mutex.

src/rshim.c
```diff
@@ -2157,9 +2228,7 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
       int rt;
 
       rshim_lock();
-      pthread_mutex_lock(&bd->mutex);
       rt = rshim_access_check(bd);
```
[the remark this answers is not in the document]
pgeng-nv: Same answer. The lock is still there. The code has been refactored. Previously this function is called directly in the timer function. Now it's called in the worker handler, which is already protected in the same bd->mutex.

src/rshim.c
pgeng-nv: * "FORCE_CMD" is a one-time command like "SW_RESET", so it can only be set to 1 and not 0.

* Apologize for didn't mention the RM number in description: https://redmine.mellanox.com/issues/3679848. You can see the use case for this feature is for BMC to take ownership from Host, kind of opposite to your example. In that case, the user will set "FORCE_CMD" to 1 to take ownership of RSHIM from host, and he/she cannot issue commands from the host for lack of access.

pgeng-nv: In my new push below, I also improved the "FORCE_CMD" misc display to avoid the perception that "0" is allowed to write. Now it's similar to "SW_RESET".

src/rshim.c
```diff
@@ -29,18 +29,54 @@
...
+#define CHECK_LOCKED_MODE_MS      100
+#define CHECK_LOCKED_MODE_TICKS   (CHECK_LOCKED_MODE_MS / RSHIM_TIMER_INTERVAL)
+
+/* Timeouts in ms */
+#define TO_RSHIM_REQ_MS 5000
+#define TO_RSHIM_ONLINE_MS 5000
+
+/* Timeouts in ticks */
+#define TO_TICKS_RSHIM_REQ (TO_RSHIM_REQ_MS / OSP_MGT_INTERVAL_MS)
+#define TO_TICKS_RSHIM_ONLINE (TO_RSHIM_ONLINE_MS / OSP_MGT_INTERVAL_MS)
```
lsun100, line 42: I couldn't find where these 4 MACROs are used... 
If needed, we could consider some meaningful name with RSHIM_XXX prefix.

src/rshim.c
```diff
@@ -29,18 +29,54 @@
 /* RShim timer interval in milliseconds. */
 #define RSHIM_TIMER_INTERVAL 1
 
+/* Intervals to check the locked mode. */
+#define CHECK_LOCKED_MODE_MS      100
+#define CHECK_LOCKED_MODE_TICKS   (CHECK_LOCKED_MODE_MS / RSHIM_TIMER_INTERVAL)
```
lsun100, line 34: Maybe RSHIM_LOCK_MODE_CHECK_MS / RSHIM_LOCK_MODE_CHECK_TICK, so naming will be more close to others

src/rshim.c
```diff
@@ -29,18 +29,54 @@
...
+static int rshim_keepalive_ticks = RSHIM_KEEPALIVE_PERIOD / RSHIM_TIMER_INTERVAL;
 
 /* Keepalive magic number. */
 #define RSHIM_KEEPALIVE_MAGIC_NUM 0x5089836482ULL
 
+/* Rshim ownership management. */
+#define OSP_MGT_INTERVAL_MS        100 /* Ownership state machine interval */
+#define OSP_MGT_INTERVAL_TICKS     (OSP_MGT_INTERVAL_MS / RSHIM_TIMER_INTERVAL)
+#define RSHIM_OWNERSHIP_REQ_MAGIC_NUM 0x4F53505F524551ULL /* OSP_REQ */
+#define RSHIM_OWNERSHIP_ACK_MAGIC_NUM 0x4F53505F41434BULL /* OSP_ACK */
```
lsun100, line 61: Consider use unified naming, such as RSHIM_OSP_xxx

src/rshim.c
```diff
@@ -29,18 +29,54 @@
...
+
+/* Pause RSHIM_DBG() here after 10 messages, then resume after 1 minute */
+#define RSHIM_DBG_PAUSE 6   /* messages limit for burst print */
+#define RSHIM_DBG_RESUME 10   /* seconds before resuming */
```
lsun100, line 65: Remove if not needed.

src/rshim.c
```diff
@@ -213,6 +250,10 @@ static uint32_t rshim_timer_interval = RSHIM_TIMER_INTERVAL;
 
 static void rshim_fifo_msg_update_checksum(rshim_tmfifo_msg_hdr_t *hdr);
 
+static int rshim_update_locked_mode(rshim_backend_t *bd);
+
+static int handle_ownership_transfer(rshim_backend_t *bd);
```
lsun100, line 255: rshim_handle_ownership_transfer()? So the naming is more consistent except for very short utility functions.

src/rshim.c
```diff
@@ -1746,7 +1787,17 @@ static void rshim_work_handler(rshim_backend_t *bd)
...
+    bd->has_locked_work = 0;
+    rshim_update_locked_mode(bd);
+  }
   pthread_mutex_unlock(&bd->mutex);
+
+  /* Ownership transfer needs to run without holding bd->mutex. */
+  if (bd->has_osp_work) {
+    bd->has_osp_work = 0;
+    handle_ownership_transfer(bd);
+  }
```
lsun100, line 1800: Need to double check. All rshim read/write should be protected by bd->mutex

src/rshim.c
```diff
@@ -2069,43 +2120,40 @@ rshim_backend_t *rshim_find_by_dev(void *dev)
   return NULL;
 }
 
+/*
+ * Must run with bd->mutex held.
+ *
+ * No need to call this function when value is set to 1. Just set bd->drop_mode
+ * to 1.
```
lsun100, line 2127: Can we remove these three lines. since we could still call this function even value is 1.

src/rshim.c
```diff
@@ -2069,43 +2120,40 @@ rshim_backend_t *rshim_find_by_dev(void *dev)
...
-    pthread_mutex_unlock(&bd->mutex);
     return -EALREADY;
   }
 
-  bd->drop_mode = 0;
-  if (bd->enable_device && bd->enable_device(bd, value ? false : true))
-    bd->drop_mode = 1;
-  else
-    bd->drop_mode = value;
+  bd->drop_mode = value;
```
lsun100, line 2140: This change removed like 2086 and could cause issue. We might need to change it back.

src/rshim.c
pgeng-nv: The two 5 seconds are to define protocol time for the ownership requester waiting for ACK, and for request recipient to waiting for the requestor to be active. Right now it's hard coded to 1 second. Will change this value and the protocol code to reflect this.

src/rshim.c
```diff
@@ -2157,30 +2203,183 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
...
 
   return 0;
 }
 
+static int read_rshim_sp1(rshim_backend_t *bd, uint64_t* sp1)
+{
+  int rc;
+
+  rc = bd->read_rshim(bd, RSHIM_CHANNEL, bd->regs->scratchpad1, sp1, RSHIM_REG_SIZE_8B);
+  if (rc) {
```
lsun100, line 2226: SP1 has well-defined format, so it's safe to add:  if (rc || RSHIM_BAD_CTRL_REG(*sp1))

src/rshim.c
```diff
@@ -2157,30 +2203,183 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
...
+
+static int check_rshim_sp1_magic(rshim_backend_t *bd, uint64_t magic,
+    bool* result)
+{
+  int rc;
+  uint64_t sp1;
+
+  rc = read_rshim_sp1(bd, &sp1);
+  if (rc) {
+    RSHIM_ERR("rshim%d failed to read sp1: %d\n", bd->index, rc);
```
lsun100, line 2242: Line 2227 above and line 2242 duplicate.  Only need to keep one.

src/rshim.c
```diff
@@ -2157,30 +2203,183 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
...
         rshim_set_drop_mode(bd, 1);
+        pthread_mutex_unlock(&bd->mutex);
       }
     }
   }
 
   return 0;
 }
 
+static int read_rshim_sp1(rshim_backend_t *bd, uint64_t* sp1)
```
lsun100, line 2221: Can we use existing naming style, i.e. rshim_read_sp1()...  Same as other new functions below.

src/rshim.c
```diff
@@ -2157,30 +2203,183 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
...
+  return write_rshim_sp1_magic(bd, RSHIM_OWNERSHIP_REQ_MAGIC_NUM);
+}
+
+static int write_rshim_magic_ownership_ack(rshim_backend_t *bd)
+{
+  RSHIM_INFO("rshim%d writing ownership ack magic 0x%llx\n", bd->index,
+      RSHIM_OWNERSHIP_ACK_MAGIC_NUM);
+
+  return write_rshim_sp1_magic(bd, RSHIM_OWNERSHIP_ACK_MAGIC_NUM);
+}
```
lsun100, line 2290: Looks like write_rshim_magic_ownership_req() and write_rshim_magic_ownership_ack() can be replaced with write_rshim_sp1_magic() directly with a comment where it's called.  The RSHIM_INFO can be added into write_rshim_sp1_magic() instead. The MAGIC number itself tells what it is, or it can be converted into string with the magic_to_str() function above.

src/rshim.c
pgeng-nv: This change is the newchange per our latest discussion:

_Liming Sun
7/2/2024 2:32 PM
In the ownership function, we can leave the drop mode as 1, but can call bd->enable_device(true)_

You suggested when we enter drop_mode, we call bd->enable_device(true). But if drop_mode is 0, the device is already enabled, so no need to call bd->enable_device() at all?

src/rshim.c
```diff
@@ -2157,30 +2203,183 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
...
+
+static int handle_ownership_transfer(rshim_backend_t *bd) {
+  int i, rt;
+  bool has_ack;
+  bool has_req;
+
+  if (bd->drop_mode) {
+    if (bd->force_cmd_pending) {
+      RSHIM_INFO("rshim%d executing Force command\n", bd->index);
```
lsun100, line 2299: This is the place to enable the device if drop_mode==1 && force_cmd_pending==1, in case pcie is unmapped.

if (bd->enable_device && !bd->enable_device(bd, true))
  return fail.
owner ship transfer

src/rshim.c
lsun100: Not here, it's another place (see comment 'This is the place to enable the device' below).

src/rshim.c
pgeng-nv: handle_ownership_transfer() calls rshim_set_drop_mode(), but previous implementation of rshim_set_drop_mode() holds the bd->mutex. That's why it cannot be inside the lock previously. This afternoon's drop_mode rewrite has removed the locking in rshim_set_drop_mode(). So It's possible to put this section back inside lock again. Will change.

src/rshim.c
pgeng-nv: Agree it's less impact that way. Will change.

src/rshim.c
pgeng-nv: Will change. BTW, it seems it should be `if (bd->enable_device && bd->enable_device(bd, true))`. Notice there should be no `!`.

pgeng-nv:
@lsun100 , all your current comments have been addressed in new push dff3cba. Again, you can review changes against master by using the "File changed" Tab.

src/rshim.c
```diff
@@ -2069,20 +2112,20 @@ rshim_backend_t *rshim_find_by_dev(void *dev)
...
 
-  pthread_mutex_lock(&bd->mutex);
   old_value = (int)bd->drop_mode;
   value = !!value;
   if (value == old_value) {
-    pthread_mutex_unlock(&bd->mutex);
     return -EALREADY;
   }
 
-  bd->drop_mode = 0;
```
lsun100, line 2085: Any reason to remove this line? It could cause problem.

pgeng-nv: added

src/rshim.c
```diff
@@ -2157,30 +2195,152 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
...
+      if (rt || !has_ack) {
+        RSHIM_WARN("rshim%d failed to receive ownership transfer ack\n",
+            bd->index);
+        bd->requesting_rshim = 0;
+        return -EIO;
+      }
+      RSHIM_INFO("rshim%d received ownership transfer ack\n", bd->index);
+
+      rt = rshim_set_drop_mode(bd, 0);
+      if (rt) {
```
lsun100, line 2295: if (rt || bd->drop_mode)

src/rshim_fuse.c
```diff
@@ -851,7 +856,9 @@ static int rshim_fuse_misc_write(struct cuse_dev *cdev, int fflags,
   } else if (strcmp(key, "DROP_MODE") == 0) {
     if (sscanf(p, "%d", &value) != 1)
       goto invalid;
+    pthread_mutex_lock(&bd->mutex);
```
lsun100, line 859: not needed

src/rshim_fuse.c
```diff
@@ -851,7 +856,9 @@ static int rshim_fuse_misc_write(struct cuse_dev *cdev, int fflags,
...
     rshim_set_drop_mode(bd, value);
+    pthread_mutex_unlock(&bd->mutex);
```
lsun100, line 861: not needed.

src/rshim_usb.c
```diff
@@ -922,10 +922,16 @@ static int rshim_usb_probe_one(libusb_context *ctx, libusb_device *usb_dev,
 
   /* Notify that device is attached. */
   rc = rshim_notify(bd, RSH_EVENT_ATTACH, 0);
-  pthread_mutex_unlock(&bd->mutex);
   if (rc)
     goto error;
```
lsun100, line 926: missed unlock for the "goto error"

src/rshim.c
```diff
@@ -2157,30 +2196,152 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
...
+  } else {
+    has_req = false;
+    rt = rshim_check_sp1_magic(bd, RSHIM_OSP_REQ_MAGIC_NUM, &has_req);
+
+    if (!rt && has_req) {
+      RSHIM_INFO("rshim%d received ownership transfer request\n", bd->index);
+
+      rshim_lock();
+
+     RSHIM_INFO("Notifying the requester with ACK\n");
```
asmaamellanox, line 2312: Please align this properly with the rest

src/rshim.c
```diff
@@ -2069,20 +2112,20 @@ rshim_backend_t *rshim_find_by_dev(void *dev)
...
 
+/*
+ * Must run with bd->mutex held.
+ * */
 int rshim_set_drop_mode(rshim_backend_t *bd, int value)
 {
   int old_value;
   int rt = 0;
 
-  pthread_mutex_lock(&bd->mutex);
```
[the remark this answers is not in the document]
pgeng-nv: Will change.

src/rshim.c
```diff
@@ -2157,30 +2195,152 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
...
 
       rshim_lock();
-      pthread_mutex_lock(&bd->mutex);
       rt = rshim_access_check(bd); 
-      pthread_mutex_unlock(&bd->mutex);
       rshim_unlock();
       if (rt) {
         RSHIM_INFO("rshim%d attached by another device. Entering Drop Mode\n",
             bd->index);
+        pthread_mutex_lock(&bd->mutex);
```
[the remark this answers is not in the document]
pgeng-nv: Will remove.

src/rshim.c
pgeng-nv: Not sure why `|| bd->drop_mode` is needed. If "rt" is non-zero, meaning the function failed to set `bd->drop_mode` from 1 to 0, which means it must be `bd->drop_mode` must be 1 already.

src/rshim.c
```diff
@@ -2157,30 +2195,152 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
...
+
+      bd->requesting_rshim = 1;
+
+      /* sending req and checking ack multiple times to ensure the transfer */
+      for (i = 0; i < 10; i++) {
+        rt = rshim_write_sp1_magic(bd, RSHIM_OSP_REQ_MAGIC_NUM);
+        if (rt) {
+          RSHIM_ERR("rshim%d failed to write ownership transfer req\n",
+              bd->index);
+          continue;
```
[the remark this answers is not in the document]
pgeng-nv: Will change.

src/rshim.c
lsun100: Yes, you're right. For this case, checking 'rt' should be good enough though checking both might be more robust (for example, if drop_mode is already 0 and we call rshim_set_drop_mode(bd, 0) again, it'll return failure. Up to you whether you want to add this change.

src/rshim.c
```diff
@@ -2092,22 +2141,23 @@ int rshim_set_drop_mode(rshim_backend_t *bd, int value)
...
       bd->drop_mode = 1;
       rt = -EACCES;
     }
-    pthread_mutex_unlock(&bd->mutex);
     rshim_unlock();
   }
+
+  if (has_bd_lock)
+    pthread_mutex_unlock(&bd->mutex);
+
```
lsun100, line 2160: Can we change these lines back? It's done in such way for a reason to avoid potential dead lock.

src/rshim.c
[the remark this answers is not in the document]
lsun100: Looks like changes in this function are not complete. We can use trylock to repalce lock and add a check when doing the unlock.

src/rshim.c
pgeng-nv: Then should it be `rt && bd->drop_mode` here?

pgeng-nv: Or how about `if (rt && rt != -EALREADY)`?

lsun100: Suggest either "if (rt)" like you did before,  or "if (bd->drop_mode)"

pgeng-nv: Use `if(rt)`, but make rshim_set_drop_mode() to return 0 when the value is already there: 1e89c98.

src/rshim.c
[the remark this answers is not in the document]
pgeng-nv: Updated rshim_update_locked_mode() with try-lock so it can be called with bd->mutex held: 2c050a1

src/rshim.c
lsun100: better not to change the locking order between rshim_lock() and pthread_mutex_unlock(), which could potentially cause deadlock.

src/rshim.c
```diff
@@ -2157,30 +2208,159 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
...
+          return -EIO;
+        }
       }
     }
   }
 
+  if (has_bd_lock)
+    pthread_mutex_unlock(&bd->mutex);
+
+  return 0;
```
lsun100, line 2230: same in this function

src/rshim.c
pgeng-nv: Changes pushed in e605c95

src/rshim.c
```diff
@@ -2157,30 +2212,159 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
       int rt;
 
       rshim_lock();
-      pthread_mutex_lock(&bd->mutex);
```
lsun100, line 2160: missed this one?

src/rshim.c
```diff
@@ -2157,30 +2212,159 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
...
+          if (has_bd_lock)
+            pthread_mutex_unlock(&bd->mutex);
+          return -EIO;
+        }
       }
     }
   }
 
+  if (has_bd_lock)
+    pthread_mutex_unlock(&bd->mutex);
```
lsun100, line 2232: still there this?

src/rshim.c
```diff
@@ -2157,30 +2212,159 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
...
+        }
+      }
+
+      usleep(RSHIM_OSP_TO_ONLINE_MS * 1000);
+
+      RSHIM_INFO("rshim%d entering drop mode\n", bd->index);
+      rt = rshim_set_drop_mode(bd, 1);
+      if (rt) {
+        RSHIM_ERR("rshim%d failed to enter drop mode\n", bd->index);
+        rshim_unlock();
```
lsun100, line 2349: Do we really need this?

pgeng-nv: Wow, great catch. Will remove.

src/rshim.c
```diff
@@ -2157,30 +2211,155 @@ static int rshim_update_locked_mode(rshim_backend_t *bd)
       int rt;
 
       rshim_lock();
-      pthread_mutex_lock(&bd->mutex);
+      has_bd_lock = !pthread_mutex_trylock(&bd->mutex);
       rt = rshim_access_check(bd); 
-      pthread_mutex_unlock(&bd->mutex);
-      rshim_unlock();
+      if (has_bd_lock)
+        rshim_unlock();
```
lsun100, line 2217: need fix here, missing pthread_mutex_unlock and changed rshim_unlock incorrectly.

pgeng-nv: Thanks. Fixed in e36d992.

src/rshim.c
```diff
@@ -2073,42 +2116,56 @@ int rshim_set_drop_mode(rshim_backend_t *bd, int value)
...
+    if (has_bd_lock)
+      pthread_mutex_unlock(&bd->mutex);
+    return 0;
   }
 
-  bd->drop_mode = 0;
-  if (bd->enable_device && bd->enable_device(bd, value ? false : true))
-    bd->drop_mode = 1;
-  else
-    bd->drop_mode = value;
```
lsun100, line 2089: We can't move this block...
Considering drop mode changed from 1 to 0, it needs to enable the device first or else the following rshim_access_check() will not work.

src/rshim.c
```diff
@@ -2073,42 +2116,56 @@ int rshim_set_drop_mode(rshim_backend_t *bd, int value)
...
+  bd->drop_mode = 0;
+  if (bd->enable_device && bd->enable_device(bd, value ? false : true))
+    bd->drop_mode = 1;
+  else
+    bd->drop_mode = value;
+
+  if (bd->drop_mode)
+    bd->drop_pkt = 1;
+  else
+    rshim_fifo_sync(bd, true);
```
lsun100, line 2163: Let's not to reorganize the code order in this commit unless there is good reason.

pgeng-nv: The quick reason is when it's called in fuse thread to set drop mode to 0 when it's currently 1, it will first set bd->drop_mode to 1 and then do access check. But once bd->drop_mode is set to 1, another thread could start writing keep alive, causing the access check to fail. That's why switch the order helps. 

This is a tentative solution that aims to prove the fix. If we can come up with a less invasive fix that'll be better

lsun100: "But once bd->drop_mode is set to 1, another thread could start writing keep alive, causing the access check to fail. " 
Do you mean the peer rshim tries to write keep alive?  It's very rare unless we're trying to bring up the peer side at the same time. Even it happens, it's expected behavior because the peer side 'tries' to take over.

This change will cause problem when switching drop_mode from 1 to 0. The rshim resource has been released when drop_mode is 1. In such case, it'll try to read/write register without rshim resource mapped.

pgeng-nv: >> Do you mean the peer rshim tries to write keep alive? 
No. I mean the main thread that writes KeepAlive magic to SP1 in the same rshim driver.

>> This change will cause problem when switching drop_mode from 1 to 0. The rshim resource has been released when drop_mode is 1. In such case, it'll try to read/write register without rshim resource mapped.
I got this part. I am trying to check why it seems doing fine in my test. And also trying to find a solution without this issue.

lsun100: The original code is fine. This block of code is done inside pthread_mutex_lock(), there is no other rshim write could be done (thus no way for the timer thread to send out keepalive).

pgeng-nv: The block of code is not inside pthread_mutex_lock but inside pthread_mutex_trylock, which could fail if the thread running keep-alive already holds the bd->lock. This race condition explains why it succeeds sometimes but not other times.

lsun100 (approved):

PR #156, opened 2024-06-30
bfb-install: enhancement for NIC mode

lsun100:
This commit contains enhancement for BFB installation in NIC mode.

- Poll for 'In Enhanced NIC mode' message as installation completion.

- Unbind mlx5_core driver before installation, bind it back once it's done.

- Disable rshim log 'CLEAR_ON_READ' after seeing the installation completion indication.

RM #3955968 #3956059

### DIFF
--- a/scripts/bfb-install
+++ b/scripts/bfb-install
@@ -371,18 +371,50 @@ push_boot_stream()
   fi
 }
 
+check_nic_mode()
+{
+  local str
+
+  [ "$mode" != "local" ] && return
+
+  # Get PCIE BDF
+  str=`cat ${rshim_node}/misc | grep DEV_NAME | awk '{print $2}'`
+  str=${str:5}
+  str=${str/.*/.0}
+  pcie_bdf="$str"
+
+  # Check NIC mode
+  str=`mlxconfig -d $pcie_bdf -e q INTERNAL_CPU_OFFLOAD_ENGINE 2>/dev/null | grep INTERNAL_CPU_OFFLOAD_ENGINE | awk '{print $(NF-1)}'`
+  if [ ."$str" = ."DISABLED(1)" ]; then
+    is_nic_mode=1
+  else
+    is_nic_mode=0
+  fi
+}
+
 # Wait for RSHIM to finish updating by monitoring keywords in the RSHIM log
 #
 # Global variables used:
 #   $mode, $rshim_node, $remote_mode
 wait_for_update_to_finish()
 {
-  local filter="Reboot|finished|DPU is ready|Linux up"
+  # 'filter0' indicates bfb installation completion, thus CLEAR_ON_READ
+  # can be disabled for next boot. 'filter' is related to specific mode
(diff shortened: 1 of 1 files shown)


scripts/bfb-install
```diff
@@ -371,18 +371,47 @@ push_boot_stream()
...
+}
+
 # Wait for RSHIM to finish updating by monitoring keywords in the RSHIM log
 #
 # Global variables used:
 #   $mode, $rshim_node, $remote_mode
 wait_for_update_to_finish()
 {
-  local filter="Reboot|finished|DPU is ready|Linux up"
+  local filter0 filter
```
pgeng-nv, line 402: Better add a comment here on why we need two filters. Like "filter" is for installation completely finished, and "filter0" is for partially finished but OK to turn off CLEAR_ON_READ?

scripts/bfb-install
```diff
@@ -508,6 +547,9 @@ trap cleanup EXIT INT TERM
 run_cmd_local_ready=0     # whether run_cmd* local functions are ready to use
 run_cmd_remote_ready=0    # whether run_cmd* remote functions are ready to use
 
+is_nic_mode=0     # Flag to indicate whether DPU in NIC mod or not
+pcie_bdf=""       # PCIE BFD
```
pgeng-nv, line 551: BFD --> BDF

pgeng-nv (approved):
+1

scripts/bfb-install
lsun100: None, added comment for it:
  # 'filter0' indicates bfb installation completion, thus CLEAR_ON_READ
  # can be disabled for next boot. 'filter' is related to specific mode
  # which might include extra configuration or booting, and is used as
  # the exit condition of the bfb-install script.

scripts/bfb-install
lsun100: done

pgeng-nv (dismissed):
I have no other comments.

scripts/bfb-install
```diff
@@ -371,18 +371,51 @@ push_boot_stream()
...
+  [ "$mode" != "local" ] && return
+
+  # Get PCIE BDF
+  str=`cat ${rshim_node}/misc | grep DEV_NAME | awk '{print $2}'`
+  str=${str:5}
+  str=${str/.*/.0}
+  pcie_bdf="$str"
+
+  # Check NIC mode
+  str=`mlxconfig -d "$str" q INTERNAL_CPU_OFFLOAD_ENGINE 2>/dev/null | grep INTERNAL_CPU_OFFLOAD_ENGINE | grep 'DISABLED(1)'`
```
vladsokolovsky, line 387: str=`mlxconfig -d "$str" -e q INTERNAL_CPU_OFFLOAD_ENGINE 2>/dev/null | grep 'DISABLED(1)'`

- This will query specific parameter and make it faster.

lsun100: Done

vladsokolovsky (approved):

PR #164, opened 2024-07-16
fix: add missing --force in help menu

pgeng-nv:
[RM 3991837](https://redmine.mellanox.com/issues/3991837)


pgeng-nv:
> Can we have a RM # here (and create one if not existing)?

Createc RM [3965694](https://redmine.mellanox.com/issues/3965694)

lsun100 (approved):

PR #166, opened 2024-07-25
bfb-install: adjust the log file to be per rshim

lsun100:
RM #4007297

### DIFF
--- a/scripts/bfb-install
+++ b/scripts/bfb-install
@@ -424,8 +424,8 @@ wait_for_update_to_finish()
   # Set display level to 2 to show more information
   run_cmd_exit $mode "echo 'DISPLAY_LEVEL 2' > ${rshim_node}/misc"
 
-  # Remove old log file.
-  rm -f ${LOG_FILE}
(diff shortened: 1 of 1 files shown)


scripts/bfb-install
```diff
@@ -778,7 +781,7 @@ fi
 # Check NIC mode and unbind mlx5_core driver in NIC mode.
 check_nic_mode
 if [ $is_nic_mode -eq 1 -a -n "$pcie_bdf" ]; then
-  echo "$pcie_bdf" > /sys/bus/pci/drivers/mlx5_core/unbind
+  echo "$pcie_bdf" > /sys/bus/pci/drivers/mlx5_core/unbind 2>/dev/null
```
pgeng-nv, line 784: Just want to make sure this line to ignore errors is intended .

pgeng-nv (approved):

scripts/bfb-install
lsun100: yes, it's intended and was requested from Moshe.

PR #167, opened 2024-07-25
Make sending the initial force command a one-time event.

pgeng-nv:
Fix a bug https://redmine.mellanox.com/issues/4005070. The bug is that when host receives a misc command of "SW_RESET", BMC rshim will re-probe, causing ownership command to be resent, since that logic is inside the probe function.

This change moves setting the pending command flag logic to `main()` function right after parsing command line and loading the configuration file. It uses a global array to store the flag. So it makes sure the `-F` command line option or `FORCE_MODE` configuration file option is consumed only once, even when bd is reallocated in probing.

RM 4005070


man/rshim.8
```diff
@@ -111,18 +111,26 @@ Run in forground.
 
 -F, --force
 .in +4n
-Put rshim driver in ownership-forceable mode (aka. force mode). In this mode, the driver will send
-a one time ownership request command to the other rshim backends upon start-up
-if the rshim interface is already attached by another backend.  It also enables
-"/dev/rshim<N>/" creation even when rshim is not attached, allowing user to
-manually send a ownership request (aka FORCE_CMD) to "/dev/rshim<N>/misc"
-interface at any time.
+Put rshim driver in ownership-forceable mode (aka. force
```
lsun100, line 114: This line seems short. Is it on purpose?

lsun100:
> https://redmine.mellanox.com/issues/4005070

Suggest removing the ' https://redmine.mellanox.com/issues/4005070', and change 'RM 4005070' to 'RM #4005070' for consistency.

src/rshim.c
```diff
@@ -220,13 +220,15 @@ bool rshim_has_pcie_reset_delay = false;
...
 int rshim_pcie_intr_poll_interval = 10;  /* Interrupt polling in milliseconds */
-int rshim_force_mode = 0;  /* Request ownship when entering drop mode */
+int rshim_force_mode = 0;  /* always keep /dev/rshim<N> & send a force cmd */
 
 /* Array of devices and device names. */
 rshim_backend_t *rshim_devs[RSHIM_MAX_DEV];
 char *rshim_dev_names[RSHIM_MAX_DEV];
 char *rshim_blocked_dev_names[RSHIM_MAX_DEV];
 
+bool rshim_force_cmd_pending[RSHIM_MAX_DEV];
```
lsun100, line 230: Add a comment for this one?

By the way, you're changing FORCE_CMD to FORCE_MODE. Would it make more sense to use 'rshim_force_mode_pending' here?  Just a suggestion, but not important.

src/rshim.c
```diff
@@ -2274,9 +2276,9 @@ static int rshim_handle_ownership_transfer(rshim_backend_t *bd)
   bool has_req;
 
   if (bd->drop_mode) {
-    if (bd->force_cmd_pending) {
-      RSHIM_INFO("rshim%d executing Force command\n", bd->index);
-      bd->force_cmd_pending = 0;
+    if (rshim_force_cmd_pending[bd->index]) {
+      RSHIM_INFO("rshim%d executing Force (REQ Ownership) command\n", bd->index);
```
lsun100, line 2280: The RSHIM_INFO will be logged. Maybe we could simplify the message a little bit? like
"rshim%d executing OwnershipReq command"?

src/rshim.c
```diff
@@ -2321,6 +2323,12 @@ static int rshim_handle_ownership_transfer(rshim_backend_t *bd)
       RSHIM_INFO("rshim%d regained ownership successfully\n", bd->index);
     }
   } else {
+    if (rshim_force_cmd_pending[bd->index]) {
+      RSHIM_INFO("Skipping Force cmd for rshim%d as it's already attached\n",
```
lsun100, line 2327: "Skipping OwnershipReq cmd for rshim%d (already attached)"?

src/rshim.c
pgeng-nv: It's different and explained in rshim.conf and rshim.8.

FORCE_MODE means 1) keeping /dev/rshim<N>/misc and 2) sending out a Force command.

Will add comment here.

src/rshim.c
pgeng-nv: I changed Force to ReqOwnership yesterday since Req Ownership is more intuitive. But QA and DOCs have been using them a lot already. So I decided to keep "Force". I will remove the "REQ Ownership" part to avoid confusion.

src/rshim.c
pgeng-nv: I had to use"cmd" to keep line width under 80 :-(. Your version seems a good one that fits in 80 characters. Will try.

pgeng-nv:
> > https://redmine.mellanox.com/issues/4005070
> 
> Suggest removing the ' https://redmine.mellanox.com/issues/4005070', and change 'RM 4005070' to 'RM #4005070' for consistency.

I thought the comments above are just for the PR not the actual git commit? Will add "#" to the Git commit.

man/rshim.8
pgeng-nv: Good catch. Will change.

lsun100 (approved):

PR #170, opened 2024-08-01
bfb-install: fix NIC_MODE installation for BF2

lsun100:
This commit fixes check_nic_mode() in the bfb-install script which is used to determination the completion of the installation and is only applicable to BF3.

RM #4005240


DLNvda (approved):

pgeng-nv (commented):
+2

PR #173, opened 2024-08-05
Fix rshim deb package for DOCA build on Ubuntu

pgeng-nv:
RM #4023279

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -2,10 +2,10 @@
 set -e
 
 case "$1" in
-  # Unmask the service if it was previously masked.
-  # rshim service will be masked after "dpkg -i" or "apt remove" before 2.0.30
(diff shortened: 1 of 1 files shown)


debian/postinst
```diff
@@ -2,10 +2,10 @@
 set -e
 
 case "$1" in
+  configure)
   # Unmask the service if it was previously masked.
   # rshim service will be masked after "dpkg -i" or "apt remove" before 2.0.30
   deb-systemd-helper unmask 'rshim.service' >/dev/null || true
```
lsun100, line 8: Can we add two more spaces for these three lines?

pgeng-nv: Good catch. Pushed.

lsun100 (approved):

PR #175, opened 2024-08-08
Add a delay to workaround mysterious boot open issue during booting

lsun100:


DLNvda (approved):

src/rshim.c
```diff
@@ -706,9 +706,18 @@ int rshim_reset_control(rshim_backend_t *bd)
...
+  time_t t;
 
   pthread_mutex_lock(&bd->mutex);
 
+  /* Add a small delay. */
+  time(&t);
+  if (!bd->register_time || difftime(t, bd->register_time) < 2) {
+    RSHIM_INFO("rshim%d: boot not ready\n", bd->index);
+    pthread_mutex_unlock(&bd->mutex);
+    return -EBUSY;
```
pgeng-nv, line 718: -EAGAIN?

pgeng-nv (approved):
+2

src/rshim.c
lsun100: Not -EAGAIN, because -EAGAIN might cause the caller to retry which is not the intent.

pgeng-nv: I thought the intent is indeed to inform the user to try again later since it'll timeout in 2 seconds, unless it's not a interactive session but a program doing a while-loop trying without delay.

lsun100: There is a mysterious open/close of /dev/rshim0/boot during boot, which caused ARM reset. The goal of this WA is to ignore this request, though more investigation should be needed to find out how this boot file is opened.

PR #180, opened 2024-08-28
Revert the 2-second delay for RM #4049579

lsun100:
This pull request contains changes to revert commit 75d0d3bef258 for RM #4049579.
It also fixes two compiling warnings.


pgeng-nv (approved):
+2

PR #196, opened 2024-10-31
bfb-script: exit when pushing bfb error

pgeng-nv:
When error happens during pushing bfb, bfb-install will now exit.It will also monitor critical error in rshim log and exit if it happens.

RM #4092766


scripts/bfb-install
```diff
@@ -197,7 +197,11 @@ push_boot_stream_via_local_rshim()
     ${BF_REG} $(basename ${rshim_node}) 0x13000318.64 0x4 >/dev/null
   fi
 
-  $sudo_prefix sh -c "cat ${bfb} ${cfg:+$cfg} ${rootfs:+${rootfs}} ${pv:+| ${pv} | cat -} > ${rshim_node}/boot"
+  if ! $sudo_prefix sh -c "cat ${bfb} ${cfg:+$cfg} ${rootfs:+${rootfs}} ${pv:+| ${pv} | cat -} > ${rshim_node}/boot"; then
+    echo "Error: Failed to push boot stream via local rshim"
+    exit 1
+  fi
```
lsun100, line 203: Have we checked the case where the 'pv' command is not available? Will it return success after BFB push?

pgeng-nv: Yes it will. If `pv` command is not available, then `pv` is set to empty at line 788. Then the portion of `${pv:+| ${pv} | cat -}` evaluates to empty string, which works fine.

(If `pv` is available, then it typically evaluates to `| /usr/bin/pv | cat -`. BTW, the `cat -` could be removed.)

lsun100 (approved):

PR #197, opened 2024-11-05
Add -k/--keep-log option to allow upgrade troubleshooting

pgeng-nv:


pgeng-nv:
@lsun100, this is per our discussion to add this option to disable CLEAR_ON_READ, in case of a need to see full rshim log. The default behavior remains the same without `-k` option.

man/bfb-install.8
```diff
@@ -5,7 +5,15 @@ bfb-install \- BFB installing script for BlueField SoC over rshim driver
 
 .SH SYNOPSIS
 .B bfb-install
--b, --bfb <bfb_file> -r, --rshim [<ip:>[:port]:]rshim<N> [-m, --remote-mode <scp|nc|ncpipe>] [-R, --reverse-nc] [-c, --config <config_file>] [-f, --rootfs <rootfs_file>] [-h, --help] [-v, --verbose]
+-b, --bfb <bfb_file> -r, --rshim [<ip:>[:port]:]rshim<N>
```
lsun100, line 8: The change tries to split the option into separate lines, but this line has two options. Is it on purpose?

pgeng-nv: These two options are mandatory options, while others are all optional.

lsun100 (approved):

PR #203, opened 2024-11-15
Add BF_MODE misc output to show DPU or NIC mode

pgeng-nv:
RM #4065145


src/rshim_fuse.c
```diff
@@ -606,6 +606,25 @@ static int rshim_fuse_misc_open(struct cuse_dev *cdev, int fflags)
...
+  "NIC mode",
+  "Reserved"
+};
+
+const char* bf3_get_bf_mode(rshim_backend_t* bd) {
+  uint64_t value;
+
+  int rc = bd->read_rshim(bd, RSHIM_CHANNEL, bd->regs->scratchpad2, &value,
+                          RSHIM_REG_SIZE_8B);
+  if (rc) {
```
lsun100, line 621: We can add "|| RSHIM_BAD_CTRL_REG(value)" in the if check. Sometimes (during reset) the rc is good but the returned value is bad.

pgeng-nv: Thanks. Added.

lsun100 (approved):

PR #208, opened 2024-11-26
Exit when backend probing failed

pgeng-nv:
RM #4184083


pgeng-nv:
Need more clarification for the RM

PR #210, opened 2024-12-04
Force to stop if stuck during systemd service stop

lsun100:
This commit has changes to force stopping the rshim process if stuck during the stopping of the systemd service.

It also added a robust change for taking the semaphore of the rshim log.

RM #4187646

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -8,6 +8,7 @@
 #include <getopt.h>
 #include <netinet/in.h>
 #include <pthread.h>
+#include <semaphore.h>
 #include <signal.h>
 #include <stdio.h>
 #include <sys/epoll.h>
@@ -245,6 +246,9 @@ bool rshim_no_net;
 int rshim_log_level = LOG_NOTICE;
 bool rshim_daemon_mode = true;
 volatile bool rshim_run = true;
+
+/* rshim stop semaphore. */
+sem_t rshim_stop_sem;
 
 static uint32_t rshim_timer_interval = RSHIM_TIMER_INTERVAL;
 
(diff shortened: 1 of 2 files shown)


src/rshim.c
```diff
@@ -2773,21 +2777,30 @@ bool rshim_allow_device(const char *devname)
   return true;
 }
 
-static void *rshim_stop_thread(void *arg)
+/* Force to kill if not able to cleanup in time. */
+static void *rshim_force_stop(void *arg)
 {
-  /* Force to kill if not able to cleanup in time. */
   sleep(3);
   kill(0, SIGKILL);
```
pgeng-nv, line 2784: Maybe adding a logging here to highlight the abnormal situation, aka, the main thread hasn't exited normally, and we are resorting to forcibly kill our own process?

pgeng-nv:
My concern is whether this brute-force method of killing our own process could cause issues if some necessary clean-ups (like for rshim FIFO or sockets) are not done. Ideally we could address the root cause of the bug and fix why rshim is locked up when being stopped by systemd. But if we are confident no resource issues would be caused by this brute force termination, then the implementation itself looks good to me.

pgeng-nv (approved):

lsun100:
> My concern is whether this brute-force method of killing our own process could cause issues if some necessary clean-ups (like for rshim FIFO or sockets) are not done. Ideally we could address the root cause of the bug and fix why rshim is locked up when being stopped by systemd. But if we are confident no resource issues would be caused by this brute force termination, then the implementation itself looks good to me.

This one is very hard to reproduce, but could happen in error cases, such as a single USB synchronous operation which could take up to 40 seconds (due to the increased default value) and there might be a loop in the code to run multiple operations before going back to main loop.  Another case is that NIC_FW in bad state and returns bogus values for rshim access over PCIe. 

The changes added in this commit is only for the "systemctl stop rshim" command, which has 6 seconds (3 + 3) delay after the  command which should be good enough for normal cases before it forces to kill the process.

As for resource, it's not a concern in Linux. All resources will be released automatically once the process is gone.

PR #211, opened 2024-12-05
USB: removes the unnecessary POLLOUT polling

lsun100:
USB output is either done in synchronous or interactive way, thus no need for POLLOUT polling. It'll save quite a bit of CPU cycle usage expecially during idle. This commit also added a check do to locked mode checking only for rshim over PCIe.

RM ##4193076


pgeng-nv (approved):
Great patch. Thanks.

PR #213, opened 2024-12-06
Add option '-e' to enable exit on bfb push errors

pgeng-nv:
This change doesn't change the default rshim behavior. When the new '-e'
    option is enabled, the following specific use cases will cause rshim to
    exit:

    * Rshim is running from host with "-d” option, and
    * Errors happen in bfb-push that are detectable by rshim, including:
      - Backend attach failure ("another backend already attached")
      - Rshim "boot" devie file open failure (like when rshim is in drop
        mode or locked mode)
      - Writing to "boot" device file failure during bfb push

    Example:

    $ sudo ./src/rshim -f -l 3 -d pcie-0000:b1:00.2  --exit-on-error
    $ sudo ./src/rshim -f -l 3 -d pcie-0000:b1:00.2  --e

    Note: Not all bfb-push errors are detectable by rshim as it’s handled
    by upper layer)

    RM #4184083

### DIFF
--- a/man/rshim.8
+++ b/man/rshim.8
@@ -3,7 +3,17 @@
 .SH NAME
 rshim \- user-space rshim driver for BlueField SoC
 .SH SYNOPSIS
-rshim [options]
+rshim
+    [-b, --backend <usb|pcie|pcie_lf>]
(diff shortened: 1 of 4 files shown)


pgeng-nv:
Not needed anymore due to reduced scope.

PR #218, opened 2025-01-08
bfb-install: Handle PF1 bind/unbind in NIC mode

pgeng-nv:
RM #4221363

### DIFF
--- a/scripts/bfb-install
+++ b/scripts/bfb-install
@@ -392,12 +392,17 @@ check_nic_mode()
   # Get PCIE BDF
   str=`cat ${rshim_node}/misc | grep DEV_NAME | awk '{print $2}'`
   str=${str:5}
-  str=${str/.*/.0}
-  pcie_bdf="$str"
+  str=${str/.*/}
+  pcie_bd="$str"
+  num_ports=$(lspci -nn -s $pcie_bd | grep -v '\[0801\]' | wc -l || { echo "Failed to query PCIe ports for $pcie_bd"; echo 0; })
+  if [ "$num_ports" -eq 2 ]; then
+    [ $verbose -eq 1 ] && echo "Detected dual-port device: ${pcie_bd}.0 and ${pcie_bd}.1"
+  fi
 
   # Check NIC mode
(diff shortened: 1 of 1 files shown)


scripts/bfb-install
```diff
@@ -392,11 +392,16 @@ check_nic_mode()
...
   str=${str:5}
-  str=${str/.*/.0}
-  pcie_bdf="$str"
+  str=${str/.*/}
+  pcie_bd="$str"
+  num_ports=$(lspci -nn -s $pcie_bd | grep -v '\[0801\]' | wc -l || echo 0)
+  if [ "$num_ports" -eq 2 ]; then
+      dual_pcie_port=1
+      [ $verbose -eq 1 ] && echo "Detected dual-port device: ${pcie_bd}.0 and ${pcie_bd}.1"
+  fi
```
lsun100, line 401: Should we add  'else' to set dual_pcie_port=0? Or else line 549 below might have issue.

scripts/bfb-install
```diff
@@ -797,8 +812,15 @@ fi
 
 # Check NIC mode and unbind mlx5_core driver in NIC mode.
 check_nic_mode
-if [ $is_nic_mode -eq 1 -a -n "$pcie_bdf" ]; then
-  echo "$pcie_bdf" > /sys/bus/pci/drivers/mlx5_core/unbind 2>/dev/null
+if [ $is_nic_mode -eq 1 -a -n "${pcie_bd}" ]; then
+  if [[ -e /sys/bus/pci/drivers/mlx5_core/${pcie_bd}.0 ]]; then
+    echo "Unbinding ${pcie_bd}.0"
+    run_cmd_exit local "echo \"${pcie_bd}.0\" > /sys/bus/pci/drivers/mlx5_core/unbind"
```
lsun100, line 818: To be more robust, we can set a variable here, such as unbind_p0=1, and check it when doing binding.
The script should only do bind when it does unbind already.

lsun100: I mean, what it is not bind at the beginning due to some reason (such as on purpose)?  The change will try to bind it and might cause issue.

lsun100 (approved):

PR #225, opened 2025-02-27
bfb-install: don't exit script when rshim install fails

jlanka-nvidia:
Exiting the script results in rshim log not printed on to the screen missing the reason why the install failed.

RM #4328011

### DIFF
--- a/scripts/bfb-install
+++ b/scripts/bfb-install
@@ -200,7 +200,7 @@ push_boot_stream_via_local_rshim()
 
   if ! $sudo_prefix sh -c "cat ${bfb} ${cfg:+$cfg} ${rootfs:+${rootfs}} ${pv:+| ${pv} | cat -} > ${rshim_node}/boot"; then
     echo "Error: Failed to push boot stream via local rshim"
-    exit 1
+    return
   fi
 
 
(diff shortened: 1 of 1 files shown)


pgeng-nv:
Could this change conflict with https://redmine.mellanox.com/issues/4092766?

jlanka-nvidia:
> Could this change conflict with https://redmine.mellanox.com/issues/4092766?
> 
> I thought this RM 4328011 means that when pushing failure happens, we still want to exit, but just need to print out rshim log before exit?

Your change already addresses the exit due to timeout with :
-  filter0="Rebooting\.\.\.|finished|DPU is ready|Linux up"
+  filter0="Rebooting\.\.\.|finished|DPU is ready|Linux up|CRITICAL ERROR"

'exit 1' was not necessary as it will kill the script and prevents rshim log from being printed. Was there any specific reason to choose 'exit 1' over 'return' ?

pgeng-nv (approved):

jlanka-nvidia:
> > Your change already addresses the exit due to timeout with :
> > ```
> > -  filter0="Rebooting\.\.\.|finished|DPU is ready|Linux up"
> > +  filter0="Rebooting\.\.\.|finished|DPU is ready|Linux up|CRITICAL ERROR"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 'exit 1' was not necessary as it will kill the script and prevents rshim log from being printed. Was there any specific reason to choose 'exit 1' over 'return' ?
> 
> I was being comprehensive, just in case any other push errors that won't show up with "CRITICAL ERROR". But since that comprehensiveness is causing additional issue. I agree with your approach.

I see. If there are other push errors that cause timeout in the future, we need to address them by updating filter0 so that grep can recognize the error and exit the script.

PR #227, opened 2025-03-03
Support PCIe device name runtime update

pgeng-nv:
Rshim PCIe device name could change at run time, like from
  `pcie-0000:88:00.2` to `pcie-0000:88:00.1`. Restarting rshim service
  could update the device name, but it will impact rshim service for
  unrelated cards. This change allows for updating the rshim PCIe device
  name for a specific device only.

  Example command:

    echo "DEV_NAME" > /dev/rshim3/misc

  Example Output:

    rshim3 updating old device name pcie-0000:88:00.1 for PCIe name change
    rshim3 scanning for PCIe device with base name pcie-0000:88:00
    rshim3 found matching device pcie-0000:88:00.2
    pcie-0000:88:00.1 disable
    pcie-0000:88:00.2 enable
    rshim3 successfully reassigned to PCIe device pcie-0000:88:00.2

  RM #4244145


src/rshim_pcie.c
```diff
@@ -1168,6 +1168,80 @@ static int rshim_pcie_enable(rshim_backend_t *bd, bool enable)
   return rc;
 }
 
+static int rshim_pcie_update_dev_name(rshim_backend_t *bd, const char *base_name)
```
lsun100, line 1171: Do we need to pass the 'base_name' here? Looks like the code can be simplified without it. The called will be simpler as well.

src/rshim_pcie.c
```diff
@@ -1168,6 +1168,80 @@ static int rshim_pcie_enable(rshim_backend_t *bd, bool enable)
...
+      found = true;
+      RSHIM_INFO("rshim%d found matching device %s.%d\n", 
+          bd->index, dev_path, pci_dev->func);
+
+      /* Disable the current device */
+      if (bd->enable_device)
+        bd->enable_device(bd, false);
+
+      /* Update device function which should be the only change */
+      dev->func = pci_dev->func;
```
lsun100, line 1217: Should we update "dev->dev" as well and release the old one?

src/rshim_fuse.c
```diff
@@ -1003,6 +1003,39 @@ static int rshim_fuse_misc_write(struct cuse_dev *cdev, int fflags,
...
+               bd->index, bd->dev_name);
+
+    /* Get current pcie base name (pcie-<domain>:<bus>:<device>).
+     * Example: pcie-0000:88.00.2 -> pcie-0000:88.00 */
+    char base_name[32] = {0};
+    char *dot = strrchr(bd->dev_name, '.');
+    size_t base_len = dot - bd->dev_name;
+    strncpy(base_name, bd->dev_name, base_len);
+    base_name[base_len] = '\0';
+
```
lsun100, line 1023: Looks like this block is not necessary.

src/rshim_fuse.c
```diff
@@ -1003,6 +1003,39 @@ static int rshim_fuse_misc_write(struct cuse_dev *cdev, int fflags,
...
+    pthread_mutex_lock(&bd->mutex);
+    rc = bd->pcie_update_dev_name(bd, base_name);
```
lsun100, line 1025: Maybe we could just call it "pcie_update_dev" since it'll follow the same logic when we run the command without any name change?

src/rshim_fuse.c
pgeng-nv: Without "base_name", how do we find the new device name during PCI iteration?

src/rshim_fuse.c
pgeng-nv: The function's intent is to update name and only name. Do you think it can be expanded for other updates other than name?

src/rshim_pcie.c
pgeng-nv: Agree that "base_name" extraction can be moved to `rshim_pcie_update_dev_name()`

src/rshim_pcie.c
```diff
@@ -1168,6 +1168,80 @@ static int rshim_pcie_enable(rshim_backend_t *bd, bool enable)
...
+      continue;
+
+    /* Generate the device path */
+    snprintf(dev_path, sizeof(dev_path), "pcie-%04x:%02x:%02x",
+        pci_dev->domain, pci_dev->bus, pci_dev->dev);
+
+    RSHIM_DBG("Checking if bd %s matches %s\n", dev_path, base_name);
+
+    /* Check if the base name matches */
+    if (strncmp(dev_path, base_name, strlen(base_name)) == 0) {
```
lsun100, line 1207: Can we do 
strncmp(dev_path, bd->dev_name, strlen(dev_path)), thus no need to pass the "base_name" at all?

src/rshim_fuse.c
lsun100: I added a comment below. Since it only update the function ID, the base_name can get from bd->dev_name, and there might no need to extract the base_name at all.

src/rshim_fuse.c
lsun100: On some SPE setup with BlueField as controller card, they run rshim on the controller card., and they need to restart rshim driver currently after some test which needs PCIe rescan. This change could help it as well to recover the rshim PCIe access, but there is no name change. It's the rescan and the new pcie device enumeration that helps.

src/rshim_pcie.c
lsun100: Maybe you can check (see the other comment below) whether we really need to extract anything?

src/rshim_pcie.c
pgeng-nv: Good catch. Will fix.

src/rshim_pcie.c
pgeng-nv: removed `base_name`

src/rshim_pcie.c
pgeng-nv: `dev->dev` won't change after hiding PF2 port.

lsun100 (approved):

PR #241, opened 2025-03-19
Add graceful handling of ctrl+c

lsun100:
This commit adds ctrl+c handling instead of printing the stack trace.

RM 4368644

### DIFF
--- a/scripts/mlx-mkbfb
+++ b/scripts/mlx-mkbfb
@@ -911,4 +911,8 @@ def main(argv):
 
 
 if __name__ == "__main__":
-    main(sys.argv)
+    try:
+        main(sys.argv)
+    except KeyboardInterrupt:
+        print("Aborted (Ctrl+C)")
+        sys.exit(1)
--- a/third_party/PLDM-unpack/fwpkg_unpack.py
+++ b/third_party/PLDM-unpack/fwpkg_unpack.py
@@ -1028,4 +1028,8 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    try:
+        main()
+    except KeyboardInterrupt:
+        print("Aborted (Ctrl+C)")
+        sys.exit(1)


third_party/PLDM-unpack/fwpkg_unpack.py
```diff
@@ -931,7 +931,7 @@ def prepare_records_json(self):
         return True, json_string
 
 
-def main():
+def main(argv):
```
pgeng-nv, line 934: `argv` seems not used?

scripts/mlx-mkbfb
```diff
@@ -911,4 +911,8 @@ def main(argv):
 
 
 if __name__ == "__main__":
-    main(sys.argv)
+    try:
+        main(sys.argv)
```
pgeng-nv, line 915: `sys.argv` seems not used?

third_party/PLDM-unpack/fwpkg_unpack.py
lsun100: Without this argv, some python version reports error like:
Traceback (most recent call last):
  File "/sbin/fwpkg_unpack.py", line 1032, in <module>
    main(sys.argv)
TypeError: main() takes 0 positional arguments but 1 was given

lsun100: same usage is in mlx-mkbfb, which has "def main(argv)" to avoid such issue.

scripts/mlx-mkbfb
lsun100: I see what you mean.... yes, we could remove it. Updated it.

third_party/PLDM-unpack/fwpkg_unpack.py
lsun100: updated.

pgeng-nv (approved):

PR #242, opened 2025-03-20
bfb-install: register usage cleanup

lsun100:
This commits check chip version to define the register offsets instead of using hardcoded value. So the same script could work for both BF2 and BF3.


scripts/bfb-install
```diff
@@ -509,9 +520,28 @@ wait_for_update_to_finish()
   run_cmd_exit $mode "echo 'CLEAR_ON_READ 0' > ${rshim_node}/misc"
 }
 
+check_chip_version()
```
pgeng-nv, line 523: I feel this name implies it just "checks" the version, but in reality it actually updates variables. Maybe change the name to something like "apply_chip_specific_settings()", or "configure_chip_settings()"?

lsun100: good suggestion, will update

pgeng-nv (approved):

PR #244, opened 2025-03-22
Reduce CPU utilization by network optimization

pgeng-nv:
Increase network polling interval in low traffic to reduce CPU utilization. This resolves an issue where rshim's CPU utilization on BMC would increase significantly (e.g., from 5% to 16%) due to background CPU load.

RM #4271127

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -33,6 +33,9 @@
 /* RShim timer interval in milliseconds. */
 #define RSHIM_TIMER_INTERVAL 1
 
+/* Network polling interval in timer ticks when idle (no network activity) */
+#define RSHIM_NET_POLL_IDLE_INTERVAL_TICKS 10
+
 /* Intervals to check the locked mode. */
 #define RSHIM_CHECK_LOCKED_MODE_MS      100
 #define RSHIM_CHECK_LOCKED_MODE_TICKS   (RSHIM_CHECK_LOCKED_MODE_MS / RSHIM_TIMER_INTERVAL)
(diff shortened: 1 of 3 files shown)


src/rshim_net.c
```diff
@@ -357,7 +359,7 @@ void rshim_net_rx(rshim_backend_t *bd)
                             total_len - bd->net_rx_len,
                             TMFIFO_NET_CHAN, true);
       if (len <= 0)
-        return;
+        return data_processed;
       bd->net_rx_len += len;
```
lsun100, line 363: Should we add "data_processed = 1" here as well like line 348?

src/rshim_net.c
```diff
@@ -396,4 +400,6 @@ void rshim_net_tx(rshim_backend_t *bd)
     if (written > 0)
       bd->net_tx_len += written;
```
lsun100, line 401: We can add "data_processed = 1;" as well within the "if (written > 0) {" block.

src/rshim.c
```diff
@@ -2474,8 +2478,16 @@ static void rshim_timer_run(void)
...
       if (bd->net_fd >= 0) {
-        rshim_net_tx(bd);
-        rshim_net_rx(bd);
+        if (rshim_timer_ticks % bd->net_poll_interval == 0) {
+          tx_activity = rshim_net_tx(bd);
+          rx_activity = rshim_net_rx(bd);
+
+          /* Adjust polling interval based on activity */
+          if (tx_activity || rx_activity)
+            bd->net_poll_interval = RSHIM_TIMER_INTERVAL;
```
lsun100, line 2487: "bd->net_poll_interval = 1" seems more accurate, since this function is called at RSHIM_TIMER_INTERVAL interval already.

src/rshim_net.c
pgeng-nv: It looks to me `data_processed` must be 1 already if it reaches the second while loop. The first while loop in 341 tries to read a non-zero length of data (`sizeof(pkt-hdr)`), and it either fails and returns, or will set data_processed to 1.

src/rshim_net.c
pgeng-nv: Same as previous reply, `data_processed` will have to be 1 already when the code reaches here.

src/rshim.c
pgeng-nv: Good catch. Thanks. I changed it to 1. I also changed the initialization to 1 as well. I also renamed the variables adding "ticks" postfix to make it easier to understand.

src/rshim_net.c
lsun100: same as above.

src/rshim_net.c
pgeng-nv: Make sense. Updated.

src/rshim_net.c
pgeng-nv: Updated.

lsun100 (approved):

PR #249, opened 2025-04-02
bfb-install: cleanup the TMP_DIR

lsun100:
This commit fixes the duplicate trap action that the first definition is overwritten by the second one, thus causing TMP_DIR not cleaned up properly.

RM #4389249

### DIFF
--- a/scripts/bfb-install
+++ b/scripts/bfb-install
@@ -28,8 +28,6 @@
 # either expressed or implied, of the FreeBSD Project.
 
 TMP_DIR=$(mktemp -d)
-trap "rm -rf $TMP_DIR" EXIT
-trap "rm -rf $TMP_DIR; exit 1" TERM INT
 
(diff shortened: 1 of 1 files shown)


scripts/bfb-install
```diff
@@ -28,8 +28,7 @@
 # either expressed or implied, of the FreeBSD Project.
 
 TMP_DIR=$(mktemp -d)
-trap "rm -rf $TMP_DIR" EXIT
-trap "rm -rf $TMP_DIR; exit 1" TERM INT
+trap cleanup EXIT INT TERM
```
pgeng-nv, line 31: At this point, cleanup isn't defined. So if a trap is triggered after this line and before cleanup line, there will be an error of cleanup undefined. Suggest move this line to its original location, or move cleanup before this line.

lsun100: done

pgeng-nv (approved):

PR #254, opened 2025-05-02
Fix build error while not add syslog

essencelite:


pgeng-nv:
@essencelite, can you share a description for the errors and the fix?

PR #260, opened 2025-05-16
Added "flat" BFB support

vladsokolovsky:


lsun100 (commented):
Just a small command:
    OUTPUT_DIR=${OUTPUT_DIR:-"/tmp"}

Looks like it's created under /tmp if not specified. It means that if we run it two times, the second time it might use leftover files from last time.

Should we use 'mktemp' instead and clean the directory at the end?

lsun100 (approved):

PR #265, opened 2025-05-22
bfb-install: add command syntax for lfwp

lsun100:
This commit adds command syntax for lfwp.

-l | --lfwp: enable lfwp flow
-a | --activate: activate lfwp

If only '--lfwp' is specified, it'll do activation by default unless '--activate=0' is specified.

RM #4449636

### DIFF
--- a/man/bfb-install.8
+++ b/man/bfb-install.8
@@ -6,10 +6,12 @@ bfb-install \- BFB installing script for BlueField SoC over rshim driver
 .SH SYNOPSIS
 .B bfb-install
 -b, --bfb <bfb_file> -r, --rshim [<ip:>[:port]:]rshim<N>
+[-a, --activate <0 | 1>]
 [-c, --config <config_file>]
 [-f, --rootfs <rootfs_file>]
 [-h, --help]
 [-k, --keep-log]
+[-l, --lfwp]
 [-m, --remote-mode <scp|nc|ncpipe>]
 [-R, --reverse-nc]
 [-u, --runtime]
@@ -18,8 +20,8 @@ bfb-install \- BFB installing script for BlueField SoC over rshim driver
 
 .SH DESCRIPTION
 
-bfb-install is a utility script to install BFB images on BlueField SoC over the
-rshim driver.
+bfb-install is a utility script to install or activate BFB images on
+BlueField SoC over the rshim driver.
 
 When the "--rshim" option doesn't provide an "ip" argument, the script will run
 in local mode and try to access the local rshim device and install the BFB image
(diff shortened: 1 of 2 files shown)


scripts/bfb-install
```diff
@@ -624,6 +626,8 @@ num_bfb=0
 num_rshim=0
 max_bfb=1
 max_rshim=1
+lfwp=0
+activate=
```
vladsokolovsky, line 630: Shouldn't activate=1 by default as mentioned above?

lsun100: Updated, removed the "default=1" in command help message.

vladsokolovsky (approved):

scripts/bfb-install
```diff
@@ -1013,16 +1017,44 @@ if [ ${nic_mode} -eq 1 -a -n "${pcie_bd}" -a ${runtime} -eq 0 ]; then
...
 
-push_boot_stream
+# Push BFB and wait for result.
+if [ -e "${bfb}" ]; then
+  push_boot_stream
+  wait_for_update_to_finish
+fi
+
+# LFWP activation on PCIe host.
+if [ ${lfwp} -eq 1 -a ${activate} -eq 1 ]; then
```
pgeng-nv, line 1037: Is checking $lfwp redundant here, as we set activate to whatever lfwp value?

Could "activate" be independent of "lfwp", aka, we could activate it in a future non-lfwp case?

scripts/bfb-install
lsun100: We need to differentiate the following two cases.  Set default value 0 will make it hard.
--lfwp   ===> indicate activate=1
--lfwp activate=0

man/bfb-install.8
```diff
@@ -41,6 +43,13 @@ particularly useful for environments with firewall restrictions on the local
 host side.
 
 .SH OPTIONS
+.TP
+-a, --activate
+
+This argument is used for upgrade activation. For now it's only
+used with LFWP (Live-Firmware-Patch). '--lfwp' enables LFWP and does
+activation by default unless '--activate=0' is specified.
+
```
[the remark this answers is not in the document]
lsun100: done.
$ git diff bfb-install.8
diff --git a/man/bfb-install.8 b/man/bfb-install.8
index ca1e486..f84c336 100644
--- a/man/bfb-install.8
+++ b/man/bfb-install.8
@@ -20,8 +20,8 @@ bfb-install \- BFB installing script for BlueField SoC over rshim driver

 .SH DESCRIPTION

-bfb-install is a utility script to install BFB images on BlueField SoC over the
-rshim driver.
+bfb-install is a utility script to install or activate BFB images on
+BlueField SoC over the rshim driver.

scripts/bfb-install
lsun100: It's not the same. "set activate to whatever value lfwp has" only when 'activate' is not set. That's why activate doesn't have an initial value (empty).

pgeng-nv (approved):

PR #276, opened 2025-06-07
remove wrong use of 'struct termio'

michich:
The Linux implementation of TCGETS in rshim_fuse_console_ioctl() uses sizeof(struct termio), but the type of bd->cons_termios is 'struct termios', not 'struct termio'.
Also, "man TCGETS" confirms the expected type is 'struct termios':
  int ioctl(int fd, TCGETS, struct termios *argp);

I noticed this bug only because glibc recently removed[1] the definition of struct termio. rshim then failed to build on Fedora Rawhide.

[1] "linux/termio: remove <termio.h> and struct termio"
    https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=e04afb71771710cdc6025fe95908f5f17de7b72d

### DIFF
--- a/src/rshim_fuse.c
+++ b/src/rshim_fuse.c
@@ -389,11 +389,11 @@ static void rshim_fuse_console_ioctl(fuse_req_t req, int cmd, void *arg,
   switch (cmd) {
   case TCGETS:
     if (!out_bufsz) {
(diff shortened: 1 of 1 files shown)


lsun100 (commented):
Somehow this change caused the 'screen' app to crash when connecting to the device. We'll try to investigate it as well what else might be needed.

PR #286, opened 2025-08-15
Fix rshim drop mode driver binding

pgeng-nv:
RM #4560999


src/rshim_pcie.c
```diff
@@ -1446,6 +1459,7 @@ int rshim_pcie_init(void)
   if (rshim_pcie_has_vfio()) {
     rshim_pcie_mmap_mode = RSHIM_PCIE_MMAP_VFIO;
     rshim_sys_pci_path = SYS_VFIO_PCI_PATH;
+    rshim_pcie_register_device_ids(SYS_VFIO_PCI_PATH);
```
lsun100, line 1462: This change might break the following fall-back logic:

1. started with vfio;
2. vfio mmap() failed due to various reasons (like vfio not enabled in the kernel commandline);
3. fall-back to VFIO;
4. if VFIO still fails, fall-back to direct-mapping.

pgeng-nv: I assume the "VFIO" in 3 and 4 meant "UIO". It seems to me the change keeps the mode selection logic you listed. Can you clarify the failing path?

pgeng-nv: Fixed via 60aea49.

lsun100 (approved):

PR #294, opened 2025-09-26
bfb-install: add validation of bfb file is exist

BillZhou734:
Add the check if the input bfb file exists.

RM #4551986


lsun100 (approved):

BillZhou734:
@pgeng-nv  can you please merge it?

PR #363, opened 2026-02-20
linux: fix console handling

thillux:
Since using glibc 2.42 struct termio was no longer exposed from glibc.

An earlier fix was to provide the definition of struct termio if not given in glibc. But it mixed up the ioctls for struct termios with an internal struct termio.

This commit fixes this by providing definitions from kernel headers instead of glibc and translating between all relevant ioctl sets necessary to use screen, minicom or stty with different glibc versions.


thillux:
Do you need some further explanations or changes of this? The issue of a rshim console closing early without output is still present for us.

kmein:
We have also been running into this issue and would be happy if this got merged.

thillux:
Hi, anything further I can do/provide here for this to be merged?

thillux:
This is still broken for us. Whats your strategy with this issue?

PR #369, opened 2026-03-18
bfb-install: retry PF rebind after DPU-mode unbind

pgeng-nv:
RM #4845039 exposed that host mlx5 PFs should be unbound during a DPU-mode BFB update as well, otherwise host-side recovery and devlink access can overlap and trigger the call trace seen during long update windows.

The earlier fix did that by widening the existing NIC-mode unbind/rebind flow to DPU mode, but RM #4869241 showed that the cleanup path was still doing a one-shot bind as soon as the script exited. In DPU mode that can race with device readiness and fail with:

  echo <bdf> > /sys/bus/pci/drivers/mlx5_core/bind
  I/O error

Fix this by keeping the DPU-mode unbind behavior but making restoration readiness-tolerant:

- unbind PF0/PF1 whenever a local PCIe BDF is known and it is not a deferred update, regardless of NIC vs DPU mode
- preserve the original PF bound state and only restore PFs that were bound before bfb-install started
- move the restore logic into a helper that retries bind until success, until the device appears bound, or until a bounded timeout expires

This keeps the protection needed for RM #4845039 while preventing the issue of RM #4869241 by waiting and retrying instead of failing on an early bind attempt.

RM #4845039
RM #4869241

### DIFF
--- a/scripts/bfb-install
+++ b/scripts/bfb-install
@@ -32,6 +32,8 @@ DEBUG=${DEBUG:-0}                             # Debug mode
 BF_REG=bf-reg
 RSHIM_PIPE="/tmp/rshim_pipe"
 LOG_FILE="/tmp/bfb-install.log"
+PF_BIND_RETRY_INTERVAL=2
+PF_BIND_RETRY_TIMEOUT=180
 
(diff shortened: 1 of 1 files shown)


pgeng-nv:
Pending arch review.

PR #390, opened 2026-05-27
Software shadow ring for the DPU log scratch buffer

shihyic:
rshim: software shadow ring for the DPU log scratch buffe

#4980265

- Add a per-backend software shadow ring that mirrors the DPU's 1 KiB / 7-bit hardware log scratch buffer, fixing the install-time `Updating DPU Gol?` truncation that operators hit once the live ring saturates (~25 log lines).
- Reset the shadow on `boot-open` so each new bfb push starts with a clean log instead of inheriting the pre-install boot session's BL1/BL2 prologue.
- Add a `LOG_MSG` misc write command so host-side scripts can inject log lines without being rate-limited by the 7-bit HW ring (200-message bursts now land all 200; previously capped at ~21).
- LOG_SHADOW_SIZE  default 64 KiB.

A circular buffer in struct rshim_backend decouples producers from consumers. All shadow operations are serialized by the existing bd->mutex; no new locks introduced.

Role	  Function	            Notes
Drainer   rshim_log_shadow_tick()   Worker hook, 500 ms cadence.
                                    Pulls HW ring with force_clear=true.
                                    Gated off via is_boot_open || is_booting
                                    so it never contends with the boot push.

Host      rshim_log_shadow_msg()    Used by the new LOG_MSG command. Writes
injection                           straight to the shadow, bypasses the
                                    7-bit HW ring entirely.

Reader    rshim_log_shadow_render() Called from rshim_fuse_misc_read() at
                                    DISPLAY_LEVEL 2. Non-destructive by default;
                                    honors CLEAR_ON_READ=1 by advancing tail
                                    post-render.

Reset     rshim_log_shadow_reset()  Called from rshim_boot_open() right after
                                    is_boot_open = 1 so a fresh bfb push starts
                                    with a clean log.
Using LOG_MSG

 echo "LOG_MSG [LEVEL] <text>" > /dev/rshim<N>/misc

 - LEVEL (optional, case-sensitive): INFO | WARN | ERR | ASSERT. Defaults to INFO.

 - Module is hard-coded to MISC (host-originated); entries render as LEVEL[MISC]: <text>.

 - Free-form text up to ~1 KiB. No quoting required beyond shell rules.

Backward compatibility

 - log_shadow pointer appended at the end of rshim_backend_t. Offsets of all preceding fields (notably read_rshim / write_rshim function pointers) are unchanged - partial rebuilds against the new header don't corrupt anything.

 - LOG_SHADOW_SIZE 0 -> bd->log_shadow == NULL and every shadow function early-returns. Misc read falls back to the original rshim_log_show() with its clear-or-restore-IDX semantics. Identical behavior to the previous release.

 - CLEAR_ON_READ still controls whether the misc read is destructive; only the substrate it acts on (HW ring vs shadow) changes.

Test plan/execution results

 - Truncation fix verified on BlueField-3 / USB rshim: full bf-bundle install transcript captured from PSC BL1 START through "Installation finished / Rebooting..." , no "Updating DPU Gol" cliff.

 - Boot-open reset verified: after scp bf-bundle ... /dev/rshim0/boot, misc log starts at boot mode (rshim); pre-install boot mode (emmc) block is gone.

 - LOG_MSG 200-message burst: shadow occupancy ~10 KiB / 65536, LOG_DROPPED 0. Legacy HW-only path saturates at 21 messages.

 - Concurrent readers: bfb-install exit-pattern poller + "watch -n 5 cat /dev/rshim0/misc" both observe consistent, non-racing content.

 - LOG_SHADOW_SIZE 0: behavior, semaphore protocol, and CLEAR_ON_READ semantics identical to pre-patch baseline.

 - No scp throughput regression: side-by-side measured at ~4.6 MB/s on both pre-patch and post-patch binaries.

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -236,6 +236,21 @@ int rshim_pcie_intr_poll_interval = 10;  /* Interrupt polling in milliseconds */
 bool rshim_force_mode;                   /* Keep /dev/rshim<N> & send a force cmd */
 bool rshim_cmdmode;                      /* Command mode */
 
+/*
+ * Per-backend DPU log shadow ring capacity, in bytes.  The shadow ring
+ * is a software-side circular buffer that mirrors the DPU's 1 KiB hw
+ * scratch ring, so multiple concurrent misc readers (bfb-install +
+ * watch + redfish) can all see the same log content without racing
+ * each other for a destructive hw read.  Set to 0 to disable.
+ *
+ * Default 65536 ≈ ~1k log lines, plenty for a full bfb install
+ * transcript while costing ~64 KiB per attached rshim.  Override with
+ * a whitespace-separated "LOG_SHADOW_SIZE <bytes>" line in rshim.conf
+ * (same format as all other keys -- no '=' sign, the parser uses
+ * sscanf "%31s%63s").  Set <bytes> to 0 to disable the shadow.
+ */
+int rshim_log_shadow_size = 65536;
+
 /* Array of devices and device names. */
 rshim_backend_t *rshim_devs[RSHIM_MAX_DEV];
 char *rshim_dev_names[RSHIM_MAX_DEV];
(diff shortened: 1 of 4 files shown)


pgeng-nv:
For https://redmine.mellanox.com/issues/4980265 that this fix is addressing, the issue won’t be there if we use bfb-install instead of scp. bfb-install has native support to mitigate this exact issue due to limited size of rshim buffer.

So maybe instead of this extensive change, we ask the user to use bfb-install to do the bfb installation?

shihyic:
> For https://redmine.mellanox.com/issues/4980265 that this fix is addressing, the issue won’t be there if we use bfb-install instead of scp. bfb-install has native support to mitigate this exact issue due to limited size of rshim buffer.
> 
> So maybe instead of this extensive change, we ask the user to use bfb-install to do the bfb installation?

Added this to RM ticket. Let the verification team decide or maybe just RN it.

PR #402, opened 2026-06-30
Retry stalled boot writes

pgeng-nv:
A boot write can temporarily make no progress when the BlueField boot
FIFO has no space. The current path treats that zero-progress case as
-EINTR, so writing a BFB to /dev/rshim<N>/boot can fail even though the
caller was not interrupted.

Retry zero-progress writes until the FIFO drains or the existing boot
timeout expires. Keep the wait interruptible so Ctrl-C can abort a
stalled transfer, and keep releasing the backend mutex during the retry
delay so other rshim operations can continue.

RM #5110797

### DIFF
--- a/src/rshim.c
+++ b/src/rshim.c
@@ -856,7 +856,8 @@ boot_open_done:
 }
 
 int rshim_boot_write(rshim_backend_t *bd, const char *user_buffer, size_t count,
-                     int (*copy_in)(void *dest, const void *src, int count))
+                     int (*copy_in)(void *dest, const void *src, int count),
+                     int (*interrupted)(void *arg), void *interrupted_arg)
 {
   int rc = 0, whichbuf = 0, len;
   time_t tm;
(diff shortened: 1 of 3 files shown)


src/rshim.c
```diff
@@ -911,7 +911,15 @@ int rshim_boot_write(rshim_backend_t *bd, const char *user_buffer, size_t count,
...
-        rc = -EINTR;
+        /*
+         * No boot FIFO space is available yet. Keep the write pending until
+         * the peer drains space or the boot timeout expires, but don't block
+         * console/network/misc handling while waiting to retry.
+         */
+        pthread_mutex_unlock(&bd->mutex);
+        usleep(1000);
+        pthread_mutex_lock(&bd->mutex);
+        continue;
```
lsun100, line 932: One side effect I can see is that it might not be able to use "ctrl + c" to terminate the 'cat'.
Even we use "ctrl + c" to terminate the bfb-install, the rshim driver might be still looping here for 5 minutes...
So we can't push new images in the meantime which might cause confusing to user.

lsun100 (approved):
Have you considered to apply multi-thread for misc and console as well or maybe later?

pgeng-nv:
> Have you considered to apply multi-thread for misc and console as well or maybe later?

Not at this time. "misc" writes are quick so it's not worth it. "console" may benefit from this but it's better we see a concrete example like this RM where current single threaded approach is causing real issues.

PR #403, opened 2026-07-02
bfb-install: print BFB-Install deprecation notice on every run

DorShireto:
Display a deprecation note as both a header (at startup) and a footer (on exit) for every bfb-install invocation, informing users that DOCA 3.5 (July 2026 release) is the last DOCA version to support BFB-Install, and that DOCA-Installer will replace its full functionality starting with the Oct 2026 DOCA release.

RM #5061358

### DIFF
--- a/scripts/bfb-install
+++ b/scripts/bfb-install
@@ -78,6 +78,16 @@ echo_dbg() {
   fi
 }
 
+# BFB-Install deprecation notice. Printed as a header (at startup) and as a
+# footer (on exit) so it is shown for every bfb-install invocation.
+print_deprecation_note() {
+  cat >&2 <<'EOF'
+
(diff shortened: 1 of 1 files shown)


pgeng-nv (approved):

scripts/bfb-install
```diff
@@ -78,6 +78,16 @@ echo_dbg() {
   fi
 }
 
+# BFB-Install deprecation notice. Printed as a header (at startup) and as a
+# footer (on exit) so it is shown for every bfb-install invocation.
+print_deprecation_note() {
+  cat >&2 <<'EOF'
+
+***Note: DOCA 3.5 (July 2026 release) will be the last DOCA version to support BFB-Install. In the following DOCA release (Oct 2026) DOCA-Installer will replace the full functionality provided by BFB-Install, and BFB-Install will no longer be included in future DOCA packages. ***
```
vladsokolovsky, line 86: Maybe, add the full path for the doca-installer (/usr/sbin/doca-installer)?

PR #407, opened 2026-07-06
Enhance bfb-tool: Add --without-gi option for repacking without DPU and NIC firmware

Shai1-Levi:


DorShireto:
Shift from the requierment of --without-gi to expect true or false to be a stand-alone action

DorShireto (approved):

PR #409, opened 2026-07-08
Fix PCIe NIC/DPU FW reset handshake stall under VFIO/UIO

shravankumarr:
The NIC/DPU FW reset handshake between the firmware and the rshim driver is signalled via scratchpad6 and a PCIe INTx interrupt. On setups where the rshim device is bound via VFIO (and UIO), the INTx was not delivered to the interrupt eventfd, so rshim_pcie_intr() never ran. As a result the driver never wrote the ACK/START_ACK reply, the firmware timed out waiting for it, and the reset was aborted (SP6 cleared) instead of completing.

Firmware does expose the full request in scratchpad6, so the driver does not actually need the interrupt to observe it. Two changes make the handshake complete reliably:

1. Poll fallback in the interrupt thread. rshim_pcie_intr_thread() now waits on the interrupt fd with poll() and a 100ms timeout instead of a blocking read(). On timeout it checks scratchpad6 directly, so the handshake proceeds even when INTx is never delivered. The eventfd fast path is preserved when it does fire. The check runs in the per-device interrupt thread, so the START downtime sleep does not stall the shared housekeeping timer.

2. Act only on firmware state transitions. Because the handler is now also driven by polling, it can observe the same rst_state many times. The reply write is a read-modify-write of the whole SP6 word, which includes the firmware-owned (RO) rst_state field, so re-writing on every poll clobbered the firmware's transition to START and livelocked at REQUEST/ACK. A per-device rst_last_state is added so the handler responds once per state change.

Note: in VFIO/UIO mode this adds a periodic scratchpad6 read (one 8-byte register access per device every 100ms) in place of purely blocking on the eventfd.

RM #5000054


src/rshim_pcie.c
```diff
@@ -899,6 +927,7 @@ static void *rshim_pcie_intr_thread(void *arg)
 {
   rshim_pcie_t *dev = arg;
   uint8_t intr_buf[16];
+  struct pollfd pfd;
```
lsun100: Maybe struct pollfd pfd = {0}; ?

src/rshim_pcie.c
```diff
@@ -912,22 +941,37 @@ static void *rshim_pcie_intr_thread(void *arg)
...
+
     __sync_synchronize();
     if (reset_seq != dev->intr_reset_seq) {
       reset_seq = dev->intr_reset_seq;
       continue;
     }
 
-    /* Interrupt handler. */
+    /* Handle interrupt, or poll scratchpad6 on timeout. */
     rshim_pcie_intr(dev);
```
lsun100: There is a potential issue with this change:  See rshim_cmdmode.c, scratchpad-6 is re-used in bf-dump, which is ok that we don't allow NIC_FW reset during bfdump.  WIth this change, rshim_pcie_intr() is always called. When doing bfdump, it could cause issues due to updated scratchpad-6

src/rshim_pcie.c
shravankumarr: Done

src/rshim_pcie.c
shravankumarr: Fixed

pgeng-nv: Can we check PCI STATUS.INTx on timeout and call rshim_pcie_intr() only when it is asserted? This would align with the architecture’s requirement that INTx trigger SP6 processing. And it will eliminate the need for the new SP3/bfdump handling.

pgeng-nv: Regarding commit message "On setups where the rshim device is bound via VFIO (and UIO), the INTx was not delivered to the interrupt eventfd, ": are we sure about that? I think I saw INTx can work in VFIO mode or UIO mode. It's just some platform/BIOS may not handle it well. RM #4546879 is an example.

lsun100 (dismissed):
It looks good to me. We can test bfdump & the NIC_reset flow if not already.

src/rshim_pcie.c
shravankumarr: Fixed

shravankumarr: > Can we check PCI STATUS.INTx on timeout and call rshim_pcie_intr() only when it is asserted? This would align with the architecture’s requirement that INTx trigger SP6 processing. And it will eliminate the need for the new SP3/bfdump handling.

Done

pgeng-nv (approved):

PR #412, opened 2026-07-28
bfb-tool: bind-mount host device nodes into chroots

DorShireto:
Chroot working directories can end up on a filesystem mounted with 'nodev' (e.g. hardened tmpfs /tmp). cpio still creates the device-node inodes during extraction, but opening them for I/O then fails with "Permission denied", regardless of the node's permission bits. This silently breaks '>/dev/null' feature-detection idioms used by some vendored tools running inside the chroot (e.g. `which python3
>/dev/null 2>&1`), which can then fall back to code paths that may not
exist in newer tool versions.

Bind-mount /dev/null and a few other common device nodes from the host into the chroot before running anything inside it. A bind mount takes on the source mount's permissions rather than the destination's, so this works regardless of how the chroot directory itself is mounted.

RM# 5165807

### DIFF
--- a/scripts/bfb-tool
+++ b/scripts/bfb-tool
@@ -72,7 +72,44 @@ EOF
 }
 
 finish() {
+	unbind_mount_chroot_devices
 	[ -n "$WDIR" ] && rm -rf "$WDIR"
+}
+
+# Bind-mount real device nodes from the host into a chroot before running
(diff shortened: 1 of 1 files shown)


DorShireto:
@vladsokolovsky please let me know if this should be also cherry picked to other branchs.

vladsokolovsky (approved):

PR #423, opened 2026-08-19
AIVO bugs

gulaak:
Addressing AIVO security vulnerabilities found in RSHIM 2.90 release.


pgeng-nv:
Could you split this into one PR per RM? The fixes are independent and touch different areas, so separate PRs will be easier to review, test, backport, and revert. Please also retain the corresponding RM number in each commit message using the `RM #<number>` format.

PR #424, opened 2026-08-20
fix rshim_sys_pci_path error

rottlog:
Build fails on non-Linux systems like FreeBSD because of rshim_sys_pci_path not being defined.

The updated port of rshim-user-space on FreeBSD now comes with a patch that fixes this until the main project has this patched:
https://cgit.freebsd.org/ports/tree/sysutils/rshim-user-space/files/patch-src_rshim__pcie.c

Thanks.


rottlog:
Can someone review this?

PR #426, opened 2026-08-25
[AIVO][rshim 2.9.0] Use-after-free on USB hotplug disconnect

gulaak:
rshim_usb_disconnect() cancelled the outstanding read/write transfers and then immediately freed dev->intr_buf, closed dev->handle, and dropped the backend reference via rshim_deref(). libusb_cancel_transfer() only requests cancellation - it does not block until the transfer is actually cancelled. The completion callback for a cancelled transfer still runs later, asynchronously, off the same urb->user_data (dev), and dereferences intr_buf/handle - all of which disconnect() had already freed, closed, or potentially deallocated via the dropped reference.

A device unplugged while a read or write was in flight could therefore race the disconnect path against its own completion callback, freeing memory the callback then read or wrote (use-after-free), or double-closing the USB handle.

Defer the actual teardown instead of doing it inline in disconnect(): mark the device as detaching and cancel both transfers, but leave dev/intr_buf/handle/the backend reference alone. The tail of the read/write completion callbacks now checks, after dropping ringlock, whether a disconnect is pending and neither transfer is outstanding any more, finishing the teardown itself if so. disconnect() also runs this same check once after cancelling, to cover the case where nothing was actually outstanding to begin with.

RM: #5194326


pgeng-nv:
This fix is non-trivial and the issue it addresses is a very edge case. It requires a narrow timing window: the USB device must disconnect while an asynchronous TMFIFO read or write is outstanding, followed by a delayed cancellation callback after teardown has begun. This makes the issue unlikely to occur in practice.

If it does occur, the expected practical impact is that the BMC rshim service, which uses the USB backend, may crash. Because the systemd unit is configured with Restart=always, the service should restart automatically, limiting the impact to a temporary interruption.

Therefore, unless we can reproduce the bug and verify that the fix introduces no regressions or other side effects, I suggest marking the RM as “Won’t Fix” and holding off on this PR.

Added Liming.

lsun100 (commented):
How did test this change? One way I can think of is to use a script to call the USB unbind/bind randomly many times on DPU BMC to check whether rshim/usb could come up properly, with or without rshim console connected from BMC.

lsun100:
> This fix is non-trivial and the issue it addresses is a very edge case. It requires a narrow timing window: the USB device must disconnect while an asynchronous TMFIFO read or write is outstanding, followed by a delayed cancellation callback after teardown has begun. This makes the issue unlikely to occur in practice.
> 
> If it does occur, the expected practical impact is that the BMC rshim service, which uses the USB backend, may crash. Because the systemd unit is configured with Restart=always, the service should restart automatically, limiting the impact to a temporary interruption.
> 
> Therefore, unless we can reproduce the bug and verify that the fix introduces no regressions or other side effects, I suggest marking the RM as “Won’t Fix” and holding off on this PR.
> 
> Added Liming.

I suspect the extra delay might cause new issues. I think we should do thorough testing if really want to keep this change.

gulaak:
> How did test this change? One way I can think of is to use a script to call the USB unbind/bind randomly many times on DPU BMC to check whether rshim/usb could come up properly, with or without rshim console connected from BMC.

I did all normal RSHIM flows through BMC and did unbind/bind to see if it was successfully brought up again. We can discuss further and if the change is valuable to try to merge, I can investigate using a script to randomly unbind/bind while operations are going on concurrently.

PR #428, opened 2026-08-25
[AIVO][rshim 2.9.0] Misc read offset permits out-of-bounds read

gulaak:
rshim_fuse_misc_read() compared the caller-supplied size against (int)(rm->len - off) without first validating off. When off > rm->len, the subtraction underflows to a negative int which is then promoted to a huge size_t for the comparison, so the clamp is skipped and fuse_reply_buf() reads from rm->buffer + off past the end of the misc buffer.

A local user reading /dev/rshimN/misc with an offset beyond the generated content (e.g. pread() with off=0x10000) could make the daemon return bytes past the per-file misc buffer, disclosing adjacent process memory or crashing the FUSE worker.

Explicitly reject offsets at or beyond rm->len before computing rm->len - off, returning a zero-length (EOF) reply instead.

RM: #5194334


src/rshim_fuse.c
```diff
@@ -838,7 +838,10 @@ static int rshim_fuse_misc_read(struct cuse_dev *cdev, int fflags,
...
 
 #ifdef __linux__
-  if (size > (int)(rm->len - off))
+  if(off < 0 || off >= rm->len){
+    size = 0;
+  }
+  else if (size > (int)(rm->len - off))
     size = rm->len - off;
   pthread_mutex_unlock(&bd->mutex);
   fuse_reply_buf(req, rm->buffer + off, size);
```
pgeng-nv, line 847: rm->buffer + off could still be invalid location. Suggest we return early on line 843.

pgeng-nv:
> RM: #5194334
Standard format is "RM #5194334" (No ":")

> Signed-off-by: Aiden Gula <agula@nvida.com>
nvida --> nvidia

src/rshim_fuse.c
gulaak: If its an invalid location size will be transferred to zero and rm->buffer + off will not be read.

pgeng-nv: I checked libfuse source code and confirmed size 0 won't cause an out-of-bound read. So my previous suggestion is a good-to-have (passing out-of-bound array address is still undefined in C) but not a blocker.

pgeng-nv (approved):

PR #430, opened 2026-08-25
[AIVO][rshim 2.9.0] Device-controlled log format string

gulaak:
Create a new validator function to make sure malicious format specifies are not executed from device reads. Only accept format specifies that are defined by the HW interface.

RM: #5194342


pgeng-nv:
As noted in my note 12 in [RM 5194342](https://redmine.mellanox.com/issues/5194342), the RShim scratch log is populated only by trusted NVIDIA-controlled firmware/UEFI/installers. Unless we identify an ordinary-user path for writing these log records, the reported security scenario depends on treating a malicious or compromised BlueField peer as untrusted, which is not part of the current product threat model.

Thus I suggest we hold off on this PR until RM #5194342 is processed for next step.

PR #437, opened 2026-09-15
console: Support Linux termios2 ioctl compatibility

pgeng-nv:
Fedora 44 terminal clients use TCGETS2/TCSETS2, which the rshim console rejects. The older TCGETS handler also returns the smaller termio layout, while setters copy a libc-sized payload without validating the received length.

Use Linux UAPI definitions and explicit field conversion to support both ioctl families with correct buffer sizes, control characters and baud-rate settings. Preserve the raw console behavior: W/F setters store settings without draining output or discarding input. FreeBSD behavior remains unchanged.

Validation on Fedora 44 x86_64 with BF-3: real ioctl boundary/EFAULT probes, minicom, direct console, bidirectional tmfifo ping and checksum-verified file transfers passed before and after guest rshim restart. SELinux remained Enforcing; guest modules were preloaded to isolate the console fix. The daemon remained stable for 60 seconds after restart.

The tested revision was 3d49b23. This revision only adds copyright comments to that code; no build or test was rerun for the amendment. ARM runtime and combined SELinux/console validation remain separate.

RM #5266798


pgeng-nv:
Abandoning this one. Will push a new one.

